PR #7, opened 2019-09-13
Refactor history data retrieval

abrown:
@jedisct1 here are some changes I made to the UI in order to more easily understand what is going on. The majority of the lines changed are due to the re-factoring or auto-formatting but there are a few behavioral changes:
 - use `HISTORY_URL` to change the location of the history server (pretty useful for debugging)
 - use a geometric mean for calculating the average slowdown
 - add tests
Would be open to discussing more in detail on https://gitter.im (or wherever, really) to sync up on why I need some of these changes and where I was thinking of making changes next.


abrown:
> The majority of the lines changed are due to the re-factoring or auto-formatting

I should re-state this as "lines changed in the code"; the test fixture and yarn lock file add a log of extra lines.

PR #9, opened 2019-09-24
Fix UI conversion issues

abrown:
I had planned for this to be the change to support multiple branches but these fixes are necessary first.


abrown:
@jedisct1, I can't yet merge... 😄

jedisct1:
Oops, my bad.

Now you should :)

PR #11, opened 2019-10-04
Compare results from multiple runtimes

abrown:
To do so, this alters the JSON output format of sightglass and adds to the `meta` information stored in `sg-history`. A visual summary: the chart now may contain lines for each WASM runtime benchmarked and the filters are "Benchmark," "Suite," "Runtime," "Date Begin," "Date End," and "Plot all."


webui/sg-view/test/retrieval.js
```diff
@@ -1,25 +1,17 @@
 import test from 'ava';
-import {history} from './fixtures/history';
+import {history} from '../../../test/fixtures/history-output';
```
jlb6740: Does this have to be moved out from under the webui directory? I think having this at the root of sightglass will be a problem for stand-alone container. Right now the docker file just does a copy of the webui directory. It expects almost everything needed by or used by webui to be in the webui directory: https://github.com/CraneStation/sightglass/blob/50ba8340bc23c04af2a5aba91e42b88e47212f6a/Dockerfile#L3

abrown: No, this is a test fixture so it is only used in testing (I wanted to make sure sightglass, sg-history, and sg-view are all on the same page); to actually run, the webui has all the files it needs.

PR #12, opened 2019-10-17
Update index.vue to fix potential CORS violation error

jlb6740:


webui/sg-view/pages/index.vue
```diff
@@ -20,13 +20,22 @@
...
   } from "../js/retrieval";
 
+  var host_history = process.env.HISTORY_URL;
+  if (host_history === "") {
+  if (process.client) {
+    var host_history = "http://"+window.location.hostname+":8001/history";
+    } else {
+      var host_history = "http://localhost:8001/history";
+    }
+  }
```
abrown: I don't think the opening brace at 24 is closed (probably because line 25 needs to go?). I think `process.env.HISTORY_URL` is not set then it should be `undefined` or `null` so I would propose:
```
let host_history = process.env.HISTORY_URL;
if (!host_history) {
  host_history = "http://" + window.location.hostname + ":8001/history";
}
```
(Using `var` inside a block re-defines `host_history` so we need to avoid it inside the if-else; also, `let` is usually preferred nowadays).

webui/sg-view/nuxt.config.js
```diff
@@ -57,6 +57,6 @@ module.exports = {
   ** Allow overriding from environment
   */
   env: {
-    HISTORY_URL: process.env.HISTORY_URL || 'http://localhost:8001/history'
+    HISTORY_URL: process.env.HISTORY_URL
```
abrown: I think if we are not going to set a default then we can get rid of this.

webui/sg-view/pages/index.vue
jlb6740: >   if (host_history === "") {
>   if (process.client) {
>     var host_history = "http://"+window.location.hostname+":8001/history";
>     } else {
>       var host_history = "http://localhost:8001/history";
>     }
>   }

line 24? It's closed at line 30? The code you referenced looks different from the lines quoted?

webui/sg-view/nuxt.config.js
jlb6740: Ok ... wasn't sure if it was used elsewhere or wanted for some other reason. It's removed.

webui/sg-view/pages/index.vue
abrown: Oh, I see; yeah, line 25 should be indented, right?

abrown: > The code you referenced looks different from the lines quoted

I think `process.client` is always true with how this being run so I removed the `if-else`.

PR #14, opened 2019-10-22
Port runner and webui branches to master

jlb6740:
Ports the runner and webui branches to master. The runner code
has been put under a new webui_runner folder and includes the
docker support, the python driver and a pluggable infrastructure
for downloading and benchmarking the latest VM updates. An expected
workflow is to launch the webui using the provided docker scripts
and to separately run an experiment that captures these results
for immediate viewing and additionally sends the results to the
webui server (expected at localhost by default) for storing and
displaying history data.


webui/LICENSE
```diff
@@ -0,0 +1,202 @@
+                                 Apache License
```
abrown: Do we still need this? The top-level directory already has a `LICENSE-APACHE`.

webui/history.json.bak
```diff
@@ -0,0 +1,3 @@
+{
+  "history": {}
+}
```
abrown: Is this file used?

webui/sg_webui_run.sh
```diff
@@ -0,0 +1,37 @@
...
+    if [[ ! -e ${HOST_BASE_PREFIX}/${SG_HISTORY_FILE} ]]; then
+        cp ${HOST_BASE_PREFIX}/history.json ${HOST_BASE_PREFIX}/${SG_HISTORY_FILE}
+    fi
+
+    docker run -it --name=${SG_DOCKER_CONTAINER_NAME} --detach --volume  ${HOST_BASE_PREFIX}/${SG_HISTORY_FILE} \
+    -p ${SG_VIEW_PORT}:${SG_VIEW_PORT} \
+    -p ${SG_HISTORY_PORT}:${SG_HISTORY_PORT} \
+    --mount type=bind,source=${HOST_BASE_PREFIX}/sg_webui_start.sh,target=/sg_webui_start.sh \
+    --mount type=bind,source=${HOST_BASE_PREFIX}/sg_webui_stop.sh,target=/sg_webui_stop.sh \
+    --mount type=bind,source=${HOST_BASE_PREFIX}/${SG_HISTORY_FILE},target=/sg-history/history.json \
```
abrown: Couple of questions about mounting here:
 - is the reason that we need to mount the `.sh` files because we want to be able to start/stop processes inside the container without using `docker exec ...`?
 - if we mount the history file, do we also need the `--volume` up above?

webui/sg_webui_run.sh
```diff
@@ -0,0 +1,37 @@
...
+    ${SG_DOCKER_IMAGE_NAME} tail -f /dev/null
+fi
+
+echo "Launch Webui"
+docker exec ${SG_DOCKER_CONTAINER_NAME} /sg_webui_start.sh ${SG_HOSTNAME} ${SG_VIEW_PORT} ${SG_HISTORY_PORT} ${SG_HISTORY_LOCATION} &
```
abrown: I would have expected that `sg_webui_start.sh` would be run from a `CMD` or `ENTRYPOINT` line in the Dockerfile. Then you could get rid of the `tail -f /dev/null` script above and the container would have the same lifetime as the web services inside. To pass in configuration, we could use `-e` (https://docs.docker.com/engine/reference/run/#env-environment-variables) or use `ENTRYPOINT` and pass the configuration as the `CMD` part (https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact).

webui/sg_webui_start.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+# Check if directory exists
+if [ ! -d "sg-view/node_modules" ]; then
+    echo "Kill servers"
+    killall node -q
+    killall sg-history -q
+fi
+echo "VIEW PORT: $2"
```
abrown: If the lifetime of the container and the lifetime of the web services were the same, then this type of thing is no longer needed (nor is `sg_webui_stop.sh` below... it can just be a `docker kill`); I guess there would still be a need to check if one of the web services crashes and then exit so that the Docker container dies when one of the web services dies...

webui_runner/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN ln -s /usr/bin/wasm-ld-8 /usr/bin/wasm-ld
+
+ENV LD_LIBRARY_PATH=/usr/local/lib
+
+RUN curl -sS -L -O https://static.rust-lang.org/dist/rust-1.37.0-x86_64-unknown-linux-gnu.tar.gz \
+	&& tar xzf rust-1.37.0-x86_64-unknown-linux-gnu.tar.gz \
+	&& cd rust-1.37.0-x86_64-unknown-linux-gnu \
+	&& ./install.sh \
+	&& cd .. \
+	&& rm -rf rust-1.37.0-x86_64-unknown-linux-gnu rust-1.37.0-x86_64-unknown-linux-gnu.tar.gz
```
abrown: A thought: we could use an `ARG` variable with a default to factor out the `1.37.0` to a single location and make it easier to update later.

webui_runner/LICENSE
```diff
@@ -0,0 +1,202 @@
+                                 Apache License
```
abrown: Same comment as the `webui` license: do we need this one anymore?

webui_runner/benchmarks/.clang-format
```diff
@@ -0,0 +1,95 @@
+---
```
abrown: 👍 You ran `clang-format` on everything in here?

webui_runner/benchmarks/shootout/xblabla20.c
```diff
@@ -0,0 +1,247 @@
...
+#ifndef NO_WASI_SUPPORT
+    assert(BUF_SIZE >= xblabla20_KEYBYTES && BUF_SIZE >= xblabla20_NONCEBYTES);
+#endif
+    int i;
+    for (i = 0; i < ITERATIONS; i++) {
+        xblabla20_xor(buf, buf, BUF_SIZE, buf, buf);
+    }
+    BLACK_BOX(buf);
+    xblabla20_teardown(buf);
+}
```
abrown: It seems like duplicating the shootout benchmarks into this lower directory could cause some confusion (e.g. "which one is the right one to use?"). Is there a way we can get to only one source file for each benchmark? If it is possible, we could open a separate issue for that.

webui_runner/plugs/base_native/polybench/build.sh
```diff
@@ -0,0 +1,61 @@
...
+./stencils/jacobi-2d/jacobi_2d.c
+./stencils/seidel-2d/seidel_2d.c
+)
+
+# Build polybench
+mkdir -p base_native_compile/
+for cfile in ${polyfiles[@]}; do 
+    echo ${CC} ${COMMON_CFLAGS} -I utilities -I linear-algebra/kernels/atax utilities/polybench.c $cfile $PRINT_TIME -o  base_native_compile/$(basename -s .c "$cfile")
+    ${CC} ${COMMON_CFLAGS} -I utilities -I linear-algebra/kernels/atax utilities/polybench.c $cfile $PRINT_TIME -o  base_native_compile/$(basename -s .c "$cfile")
+done
```
abrown: I think a Makefile might do some of this boilerplate for us (e.g. `echo`).

webui_runner/plugs/base_native/echo_build.sh
```diff
@@ -0,0 +1 @@
+echo "Native Build"
```
abrown: Not sure I understand how this is used

webui_runner/config.inc
```diff
@@ -0,0 +1,24 @@
...
+
+# Variables used by docker scripts
+HOST_BASE_PREFIX=${HOST_BASE_PREFIX:-$(cd "$(dirname ${BASH_SOURCE:-$0})" && pwd)}
+SG_DOCKER_IMAGE_NAME="sg_runner_image"
+SG_DOCKER_CONTAINER_NAME="sg_runner"
+
+# TODO: PORTS are currently hardcoded in the docker image
+# Any change here must be consistent with the docker image
+SG_HOSTNAME=localhost
+SG_RESULTS_PORT=3010
```
abrown: It's not 8000?

webui_runner/plugs/base_native/shootout/build.sh
```diff
@@ -0,0 +1,26 @@
...
+    ${CC} ${COMMON_CFLAGS} -DSKIP_ENTRY_AND_LEND $cfile -o  $(basename -s .c "$cfile")
+done
+
+#Build implementation.so
+mkdir -p ${SCRIPT_LOC}/bin; cd ${SCRIPT_LOC}/bin
+
+echo ${CC} ${COMMON_CFLAGS} "-DWORKLOAD_LOCATION=${SCRIPT_LOC}/benchmark" -shared -o implementation.so ../wrapper.c
+${CC} -fPIC ${COMMON_CFLAGS} "-DWORKLOAD_LOCATION=${SCRIPT_LOC}/benchmark" -shared -o implementation.so ../wrapper.c
+
+cd ${ORIG_CWD}
```
abrown: You could use `pushd` and `popd` here but a `Makefile` takes care of this type of thing.

webui_runner/plugs/lucet_app/download.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+source ../../config.inc
+
+>&2 echo "lucet_app download"
+mkdir -p build; cd build
+if [ ! -d lucet ]; then 
+    git clone https://github.com/fastly/lucet.git lucet
+fi
```
abrown: So if the `lucet` directory already exists, do we need to do a `pull` or `fetch` to grab the latest code?

webui_runner/plugs/lucet_app/download.sh
```diff
@@ -0,0 +1,24 @@
...
+git submodule foreach --recursive git submodule init
+git submodule foreach --recursive git submodule update
+git checkout master
+
+>&2 echo "lucet_app build"
+WASI_SYSROOT=${WASI_SYSROOT} CLANG_ROOT=${CLANG_ROOT} CLANG=${CLANG} cargo build --release
+
+>&2 echo "lucet_app builtins build"
+cd ./lucet-builtins
+make
```
abrown: Does this script eventually execute inside a Docker container?

webui_runner/plugs/lucet_app/echo_commit.sh
```diff
@@ -0,0 +1,5 @@
+ORIG_CWD="$(pwd)"
+SCRIPT_LOC="$(dirname ${BASH_SOURCE:-$0})"
+cd ${SCRIPT_LOC}/build/lucet/
+echo $(git log -1 --pretty=format:"%h")
+cd ${ORIG_CWD}
```
abrown: I think this could be reduced to `git --git-dir="$SCRIPT_LOC/build/lucet/.git" log -1 --pretty=format:"%h"`

webui_runner/plugs/lucet_app/polybench/build.sh
```diff
@@ -0,0 +1,107 @@
...
+done
+
+
+
+# Build implementation.so
+mkdir -p ${SCRIPT_LOC}/bin; cd ${SCRIPT_LOC}/bin
+>&2 echo ${CC} ${COMMON_CFLAGS} -DWORKLOAD_LOCATION=${SCRIPT_LOC}/benchmark -DVM_LOCATION=${LUCET_APP_ROOT}/build/lucet/target/release/lucet-wasi -shared -o implementation.so ../wrapper.c 
+${CC} -fPIC ${COMMON_CFLAGS} -DWORKLOAD_LOCATION=${SCRIPT_LOC}/benchmark -DVM_LOCATION=${LUCET_APP_ROOT}/build/lucet/target/release/lucet-wasi -shared -o implementation.so ../wrapper.c 
+
+cd ${ORIG_CWD}
```
abrown: So we need to have separate build scripts for every runtime-suite combination because they each need specific build tweaks?

webui_runner/plugs/lucet_app/shootout/wrapper.c
```diff
@@ -0,0 +1,72 @@
...
+                             .global_teardown = NULL,
+                             .version         = TEST_ABI_VERSION };
+
+
+static void body_wrapper(const char *name, void *ctx)
+{
+    (void) ctx;
+    char wasm_file[MAX_WASM_FILE_NAME_SIZE];
+    snprintf(wasm_file, sizeof(wasm_file), "%s %s/%s", VM, WASM_FILE_PREFIX, name);
+    system(wasm_file);
```
abrown: If we're running everything through a system call, do we need to duplicate the `wrapper.c` files across runtimes?

webui_runner/results/send.sh
```diff
@@ -0,0 +1,21 @@
...
+if [ "$#" -ne 2 ]; then
+    echo ""
+    echo "Send data in json format to sg history UI"
+    echo ""
+    echo "Usage"
+    echo "-----"
+    echo "./send.sh <history_submit_url> <json_datafile>"
+    echo ""
+    echo "Example:"
+    echo "      ./submit_data.sh localhost:8001/submit 123456 results-latest.json"
```
abrown: `123456`: is this not needed anymore?

webui_runner/plugs/wasmtime_app/echo_author.sh
```diff
@@ -0,0 +1,5 @@
+ORIG_CWD="$(pwd)"
+SCRIPT_LOC="$(dirname ${BASH_SOURCE:-$0})"
+cd ${SCRIPT_LOC}/build/wasmtime/
+echo $(git log -1 --pretty=format:"%aN")
+cd ${ORIG_CWD}
```
abrown: Are all of these Git-extraction scripts duplicated for each runtime?

webui_runner/plugs/wasmtime_app/echo_branch.sh
```diff
@@ -0,0 +1 @@
+echo master
```
abrown: Do we want to figure this out dynamically in case someone is doing some experiments with a branch?

webui_runner/plugs/wasmtime_app/echo_repo.sh
```diff
@@ -0,0 +1 @@
+echo https://github.com/CraneStation/wasmtime
```
abrown: If things get renamed, this could change; perhaps this should also be determined dynamically?

webui_runner/sightglass
```diff
@@ -0,0 +1 @@
+../
```
abrown: Is this needed?

PR #15, opened 2019-10-25
Fix parsing of Unix timestamps from results

abrown:
Previously, the Unix timestamp in the results metadata was not parsed correctly by the UI. Now, upon receiving the results, the UI will replace each run's timestamp with a JS Date object. This change also includes some formatting fixes.


jlb6740:
The time is showing down to the milliseconds which shouldn't be useful. Can we remove this on another pull request? The fix looks good so will merge. Can iterate.

PR #19, opened 2019-11-05
Add polybench support.

jlb6740:
Adds polybench to benchmark folder


abrown:
Are there any changes to this versus http://web.cse.ohio-state.edu/~pouchet.2/software/polybench/ or https://github.com/Meinersbur/polybench?

webui_runner/benchmarks/polybench/datamining/correlation/correlation.c
```diff
@@ -0,0 +1,174 @@
...
+#pragma endscop
+
+}
+
+
+#ifdef WASM_ENTRY
+void _start(void)
+#else  
+int main(int argc, char** argv)
+#endif // WASM_ENTRY
```
jlb6740: Processor directives are added to each workload to match the entry name expected by the current VMs. Similar name change is done with shootout.

jlb6740:
@abrown Ok ... Based on our offline discussion seems the changes look reasonable. Only issue is questioning the license which I agree. Will triple check that and then will merge assuming OK.

PR #20, opened 2020-01-16
Updates rust used by the containerized runner from 1.37 to 1.40

jlb6740:
Wasmtime no longer builds. This was debug to be an error parsing
cargo.lock while building. According to https://github.com/rust-lang/cargo/pull/7579
it seems rust prior to 1.38 is now obsolete w.r.t. to changes
to the lock file format.

Also updates the url for the repositories pointed to for wasmtime and lucet.


jlb6740:
@jedisct1 Hi I didn't realize this was even sitting here. Not sure others have attempted to use this aspect of sightglass but looks like the changes this patch was going for would still be needed to get it back to a working state. Will review this.

PR #26, opened 2020-07-08
Update clang and fix wasmtime app build

jlb6740:


webui_runner/Dockerfile
```diff
@@ -1,31 +1,31 @@
...
+# https://forge.rust-lang.org/infra/other-installation-methods.html
+RUN curl -sS -L -O https://static.rust-lang.org/dist/rust-1.44.1-x86_64-unknown-linux-gnu.tar.gz \
+	&& tar xzf rust-1.44.1-x86_64-unknown-linux-gnu.tar.gz \
+	&& cd rust-1.44.1-x86_64-unknown-linux-gnu \
 	&& ./install.sh \
 	&& cd .. \
-	&& rm -rf rust-1.37.0-x86_64-unknown-linux-gnu rust-1.37.0-x86_64-unknown-linux-gnu.tar.gz
+	&& rm -rf rust-1.44.1-x86_64-unknown-linux-gnu rust-1.44.1-x86_64-unknown-linux-gnu.tar.gz
 
 RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.34.0/install.sh | bash
```
abrown, line 30: I noticed this while looking through here: if you're installing NVM then you shouldn't install NodeJS separately below.

webui_runner/Dockerfile
```diff
@@ -1,31 +1,31 @@
...
-	&& tar xzf rust-1.37.0-x86_64-unknown-linux-gnu.tar.gz \
-	&& cd rust-1.37.0-x86_64-unknown-linux-gnu \
+# https://forge.rust-lang.org/infra/other-installation-methods.html
+RUN curl -sS -L -O https://static.rust-lang.org/dist/rust-1.44.1-x86_64-unknown-linux-gnu.tar.gz \
+	&& tar xzf rust-1.44.1-x86_64-unknown-linux-gnu.tar.gz \
+	&& cd rust-1.44.1-x86_64-unknown-linux-gnu \
 	&& ./install.sh \
 	&& cd .. \
-	&& rm -rf rust-1.37.0-x86_64-unknown-linux-gnu rust-1.37.0-x86_64-unknown-linux-gnu.tar.gz
+	&& rm -rf rust-1.44.1-x86_64-unknown-linux-gnu rust-1.44.1-x86_64-unknown-linux-gnu.tar.gz
```
abrown, line 28: What do you think about adding a `RUST_VERSION` variable at the top of this file and using it throughout to make the next upgrade easier? Same idea probably applies to the other installed components.

webui_runner/plugs/wasmtime_app/shootout/build.sh
```diff
@@ -7,24 +7,26 @@ source ${SIGHTGLASS_ROOT}/config.inc
 
 WASM_ENTRY=-DWASM_ENTRY
 WASI_NO_SUPPORT="-DNO_WASI_SUPPORT"
-WASI_CFLAGS="--sysroot=${WASI_SYSROOT} --target=wasm32-unknown-wasi -ggdb -gdwarf"
+#WASI_CFLAGS="--sysroot=${WASI_SYSROOT} --target=wasm32 -ggdb -gdwarf"
```
abrown, line 10: Remove this line then?

webui_runner/Dockerfile
jlb6740: I got rid of NVM which didn't seem to install the version of node being used and doesn't imply which version of node is being installed by default. Plus NVM's purpose is to manage multiple versions of node and we only need 1 version.

webui_runner/Dockerfile
```diff
@@ -1,4 +1,5 @@
 FROM ubuntu:bionic
+ENV RUST_VERISON 1.44.1
```
abrown, line 2: typo: VERISON -> VERSION 😄

webui_runner/plugs/wasmtime_app/shootout/build.sh
```diff
@@ -7,24 +7,24 @@ source ${SIGHTGLASS_ROOT}/config.inc
...
+    ${WASM_CC} ${WASM_ENTRY} ${WASI_CFLAGS} ${WASI_NO_SUPPORT} ${COMMON_CFLAGS} $(basename -s .c "$cfile").wasm.o -o $(basename -s .c "$cfile").wasm -nostartfiles
 done
 
 #Build implementation.so
 mkdir -p ${SCRIPT_LOC}/bin; cd ${SCRIPT_LOC}/bin
 echo ${WASMTIME_APP_ROOT}
->&2 echo ${CC} ${COMMON_CFLAGS} -DWORKLOAD_LOCATION=${SCRIPT_LOC}/benchmark -DVM_LOCATION=${WASMTIME_APP_ROOT}/build/wasmtime/target/release/wasmtime -shared -o implementation.so ../wrapper.c 
-${CC} -fPIC ${COMMON_CFLAGS} -DWORKLOAD_LOCATION=${SCRIPT_LOC}/benchmark -DVM_LOCATION=${WASMTIME_APP_ROOT}/build/wasmtime/target/release/wasmtime -shared -o implementation.so ../wrapper.c 
+>&2 echo ${CC} ${COMMON_CFLAGS} -DWORKLOAD_LOCATION=${SCRIPT_LOC}/benchmark -DVM_LOCATION=${WASMTIME_APP_ROOT}/build/wasmtime/target/release/wasmtime -shared -o implementation.so ../wrapper.c  -fPIC
+${CC} -fPIC ${COMMON_CFLAGS} -DWORKLOAD_LOCATION=${SCRIPT_LOC}/benchmark -DVM_LOCATION=${WASMTIME_APP_ROOT}/build/wasmtime/target/release/wasmtime -shared -o implementation.so ../wrapper.c  -fPIC
```
abrown, line 28: Looking at this again, -fPIC is in there twice (at the beginning and the end); perhaps we should move it to COMMON_CFLAGS?

webui_runner/Dockerfile
```diff
@@ -8,29 +9,26 @@ RUN apt-get update \
 
 RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|apt-key add -
 
-RUN add-apt-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-8 main"
+RUN add-apt-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main"
 
 RUN apt-get update \
 	&& apt-get install -y --no-install-recommends \
-	clang-8 lldb-8 lld-8
+	clang-10 lldb-10 lld-10
```
abrown, line 16: I think we should just merge this PR but if you're in this file again you could add `ENV CLANG_VERSION 10`

abrown:
For the future: I really wish this were running in CI; it's very difficult to know what effect these changes will have without downloading the branch and running the thing on my system. There's no reason that GitHub actions couldn't do this, right?

jlb6740:
> For the future: I really wish this were running in CI; it's very difficult to know what effect these changes will have without downloading the branch and running the thing on my system. There's no reason that GitHub actions couldn't do this, right?

Yeah agree ... for testing before merging? We should probably open an issue for that.

PR #27, opened 2020-07-08
Fix lucet build

jlb6740:


webui_runner/Dockerfile
```diff
@@ -1,31 +1,34 @@
...
 RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|apt-key add -
 
-RUN add-apt-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-8 main"
+RUN add-apt-repository "deb http://apt.llvm.org/bionic/ llvm-toolchain-bionic-10 main"
 
 RUN apt-get update \
 	&& apt-get install -y --no-install-recommends \
-	clang-8 lldb-8 lld-8
+	clang-10 lldb-10 lld-10
+#RUN curl -sL https://github.com/CraneStation/wasi-sdk/releases/download/wasi-sdk-6/libclang_rt.builtins-wasm32-wasi-6.0.tar.gz  | tar x -zf - -C /usr/lib/llvm-8/lib/clang/8.0.1
```
jedisct1, line 16: Maybe commented out code should be removed instead?

jlb6740: Hi ... yes! Sorry. Should have double checked for comments.

abrown:
This depends on #26, right? Let's merge that one first?

webui_runner/plugs/lucet_app/shootout/build.sh
```diff
@@ -8,10 +8,11 @@ source ${SIGHTGLASS_ROOT}/config.inc
 
 WASM_ENTRY=-DWASM_ENTRY
 WASI_NO_SUPPORT="-DNO_WASI_SUPPORT"
-WASI_CFLAGS="--sysroot=${WASI_SYSROOT} --target=wasm32-unknown-wasi"
- 
+#WASI_CFLAGS="--sysroot=${WASI_SYSROOT} --target=wasm32-unknown-wasi"
```
abrown, line 11: Here and below: remove comments?

webui_runner/plugs/lucet_app/shootout/build.sh
```diff
@@ -23,37 +24,20 @@ cp -r ${SIGHTGLASS_ROOT}/benchmarks/shootout/* .
...
 done
 
-
-
 #Build implementation.so
 mkdir -p ${SCRIPT_LOC}/bin; cd ${SCRIPT_LOC}/bin
-echo ${CC} ${COMMON_CFLAGS} -DWORKLOAD_LOCATION=${SCRIPT_LOC}/benchmark -DVM_LOCATION=${LUCET_APP_ROOT}/build/lucet/target/release/lucet-wasi -shared -o implementation.so ../wrapper.c 
-${CC} -fPIC ${COMMON_CFLAGS} -DWORKLOAD_LOCATION=${SCRIPT_LOC}/benchmark -DVM_LOCATION=${LUCET_APP_ROOT}/build/lucet/target/release/lucet-wasi -shared -o implementation.so ../wrapper.c 
+echo ${CC} ${COMMON_CFLAGS} -DWORKLOAD_LOCATION=${SCRIPT_LOC}/benchmark -DVM_LOCATION=/opt/lucet/bin/lucet-wasi -shared -o implementation.so ../wrapper.c
+${CC} -fPIC ${COMMON_CFLAGS} -DWORKLOAD_LOCATION=${SCRIPT_LOC}/benchmark -DVM_LOCATION=/opt/lucet/bin/lucet-wasi -shared -o implementation.so ../wrapper.c
```
abrown, line 41: Why not use the variable? Or is this script only intended to be used from within a Docker container?

webui_runner/plugs/lucet_app/download.sh
```diff
@@ -14,10 +14,7 @@ git reset --hard
 git submodule init && git submodule update
 git submodule foreach --recursive git submodule init
 git submodule foreach --recursive git submodule update
+git pull
```
abrown, line 17: Is this necessary? If so, we should document why. Also, are lines 15-16 also necessary or could a `--recursive` on line 14 be used instead?

jlb6740: Agree, there is some redundancy here .. though redundancy is short-circuited (things aren't really updated twice) . Initially, I had a more minimal set of instructions (probably just line 14 with --recursive as you point out) but I found it did not always work to put the code in the state I expected. I assumed I didn't understand some nuance here but doing things redundantly with foreach seemed to work which was the only goal at the time. I'll try to remove line 14 here.

jlb6740: Note, like some other suggestions because templates are mostly identical this change isn't specific to getting lucet to build. Probably should be in a different patch addressing all the git pulls.

webui_runner/plugs/lucet_app/shootout/build.sh
jlb6740: Yes, all of these webui_runner files are written for the container .. but this still should run outside of the container. That said, LUCETC was already defined so no reason can't have a LUCET variable as well so I've added it.

webui_runner/plugs/lucet_app/download.sh
abrown: Ok, I opened https://github.com/bytecodealliance/sightglass/issues/32 for that.

PR #28, opened 2020-07-13
Enable wamr app aot

jlb6740:


webui_runner/plugs/wamr_app_aot/echo_author.sh
```diff
@@ -0,0 +1,5 @@
+ORIG_CWD="$(pwd)"
+SCRIPT_LOC="$(dirname ${BASH_SOURCE:-$0})"
+cd ${SCRIPT_LOC}/build/lucet/
```
abrown, line 3: `build/lucet` is probably a copy-paste typo; why don't we just have the author script in one place?

webui_runner/plugs/wamr_app_aot/echo_branch.sh
```diff
@@ -0,0 +1 @@
+echo main
```
abrown, line 1: To avoid the hard-code, could this be read out of the downloaded Git directory using `git branch --show-current`?

webui_runner/plugs/wamr_app_aot/echo_author.sh
abrown: And I guess the same idea applies to some of the other scripts as well: echo_commit.sh, echo_branch.sh, etc.

webui_runner/plugs/wamr_app_aot/echo_repo.sh
```diff
@@ -0,0 +1 @@
+echo https://github.com/fastly/lucet/
```
abrown, line 1: Also a likely copy-paste error: what about `git config --get remote.origin.url`?

webui_runner/plugs/wamr_app_aot/shootout/build.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+ORIG_CWD="$(pwd)/"
+SCRIPT_LOC="$(realpath $(dirname ${BASH_SOURCE:-$0}))"
+LUCET_APP_ROOT="${SCRIPT_LOC}/.."
```
abrown, line 5: Variable names...

webui_runner/plugs/wamr_app_aot/shootout/build.sh
```diff
@@ -0,0 +1,38 @@
...
+cp -r ${SIGHTGLASS_ROOT}/benchmarks/shootout/* .
+ln -s ${WAMRC} wamrc
+
+#Build shootout
+for cfile in ./*.c; do
+    echo ${WASM_CC} ${WASI_NO_SUPPORT} ${COMMON_CFLAGS} -c $cfile -o $(basename -s .c "$cfile").wasm.o
+    ${WASM_CC}  ${WASI_NO_SUPPORT} ${COMMON_CFLAGS} -c $cfile -o $(basename -s .c "$cfile").wasm.o
+
+    echo ${WASM_CC} ${WASI_NO_SUPPORT} ${COMMON_CFLAGS} $(basename -s .c "$cfile").wasm.o -o $(basename -s .c "$cfile").wasm
+    ${WASM_CC} ${WASI_NO_SUPPORT} ${COMMON_CFLAGS} $(basename -s .c "$cfile").wasm.o -o $(basename -s .c "$cfile").wasm
```
abrown, line 27: Why create the .o file separately? Not saying we shouldn't but we might want to explain why.

webui_runner/sg_runner.py
```diff
@@ -20,6 +20,7 @@
                                     lucet_app - https://github.com/fastly/lucet (latest via git pull)
                                     wasmtime_app - https://github.com/CraneStation/wasmtime (latest via git pull)
                                     node - https://github.com/nodejs/node (latest via nvm)
+                                    wamr_app_aot - https://github.com/bytecodealliance/wasm-micro-runtime (latest via nvm)
```
abrown, line 23: Remove `(latest via nvm)`?

webui_runner/plugs/wamr_app_aot/shootout/wrapper.c
```diff
@@ -0,0 +1,70 @@
...
+
+TestsConfig tests_config = {.global_setup = NULL,
+                            .global_teardown = NULL,
+                            .version = TEST_ABI_VERSION};
+
+static void body_wrapper(const char *name, void *ctx)
+{
+    (void)ctx;
+    char wasm_file[MAX_WASM_FILE_NAME_SIZE];
+    snprintf(wasm_file, sizeof(wasm_file), "%s %s/%s", VM, WASM_FILE_PREFIX, name);
```
abrown, line 35: Don't you want this behind a flag or something to avoid polluting the results with IO?

abrown: Oh.... my mistake: I didn't read closely enough: I thought this was actually printing something. But I guess the same concern applies: we could be concatenating these strings in the `BODY` macro so the concatenation is static instead of dynamic?

abrown:
Depends on #26

webui_runner/plugs/wamr_app_aot/echo_author.sh
jlb6740: 👍 I'm guessing at the time we started with one and just did a copy exact. Also not sure if everything used git at the time .... there were about 6 or 7 VMs total at one point. But I agree, it is a good idea to share as much of this scripting as possible. The one challenge probably that the build folder will have a unique name but that can be figured out. Can we do this in a follow-up patch? It's logically separate from what this patch is trying to accomplish and affects more than just wamr.

webui_runner/plugs/wamr_app_aot/echo_branch.sh
jlb6740: I'm guessing this was a quick hack that was never implemented.

jlb6740: Logically a different patch. Can address in another patch that targets all VMs?

webui_runner/plugs/wamr_app_aot/shootout/build.sh
jlb6740: 👍 Just using a template based on an earlier version of clang and lucet that required this two step process. I reworked the flags and was able to compile right to wasm that included the correct entry points.

webui_runner/plugs/wamr_app_aot/shootout/wrapper.c
jlb6740: Not sure I understand the suggestion here? What would go in the Body macro and what would be static instead of dynamic? This template, the way this is written, is from the baseline shootout template so it's not specific to this wamr file.

jlb6740:
I've addressed all comments. The ones that I didn't address via the patch update I think are improvements and will submit new patches.

abrown:
I looked at #27 first since this contains those changes.

webui_runner/plugs/wamr_app_aot/echo_author.sh
abrown: This can get fixed in https://github.com/bytecodealliance/sightglass/issues/32.

webui_runner/plugs/wamr_app_aot/echo_branch.sh
abrown: https://github.com/bytecodealliance/sightglass/issues/32

webui_runner/Dockerfile
```diff
@@ -5,7 +5,12 @@ RUN apt-get update \
 	&& apt-get install -y --no-install-recommends \
 	apt-utils build-essential gpg-agent \
 	curl ca-certificates wget software-properties-common \
-	psmisc lsof git cmake nano
+	psmisc lsof git nano zlib1g-dev libedit-dev
+
+RUN wget -qO - https://apt.kitware.com/keys/kitware-archive-latest.asc | apt-key add -
+RUN apt-add-repository 'deb https://apt.kitware.com/ubuntu/ bionic main'
+RUN apt-get update
+RUN apt-get install --yes cmake
```
abrown, line 13: If this is to get a newer version of Cmake, why not update the base Docker image (ubuntu:bionic --> ubuntu:?); if not, maybe we should document why this requires a non-standard repo.

webui_runner/plugs/wamr_app_aot/download.sh
```diff
@@ -0,0 +1,33 @@
...
+    git clone https://github.com/bytecodealliance/wasm-micro-runtime.git wamr
+fi
+cd wamr
+git clean -fd
+git submodule foreach --recursive git clean -fd
+git reset --hard
+git submodule init && git submodule update
+git submodule foreach --recursive git submodule init
+git submodule foreach --recursive git submodule update
+git pull
```
abrown, line 17: Lines 8-17 are going to change, right?

webui_runner/plugs/wamr_app_aot/echo_commit_msg.sh
```diff
@@ -0,0 +1,5 @@
+ORIG_CWD="$(pwd)"
+SCRIPT_LOC="$(dirname ${BASH_SOURCE:-$0})"
+cd ${SCRIPT_LOC}/build/lucet/
```
abrown, line 3: These paths seem incorrect (see other similar scripts).

webui_runner/plugs/wamr_app_aot/shootout/wrapper.c
abrown: Something like this (untested): 
```
#define BODY(NAME) \
    void NAME##_body(void *ctx) { system(#VM_LOCATION " " #WORKLOAD_LOCATION "/" #NAME ".wasm.aot", ctx); }
```

It would reduce the amount of code and completely avoid any slight overhead with `snprintf`. More info [here](https://gcc.gnu.org/onlinedocs/cpp/Concatenation.html).

webui_runner/Dockerfile
jlb6740: Yes, bionic does not contain the version of cmake needed. I frankly forget which build component complained but I think it was WAMR. It's not uncommon to have to download a newer copy of cmake from a repository and that is the case here. Bionic has been the latest longterm release that most of these VMs are tested on and WASI SDK has an installer which was used, but 20.04 was released in June. I don't know what cmake comes default with 20.04 but if it is recent enough we can remove this here. I will add a comment.

webui_runner/plugs/wamr_app_aot/download.sh
jlb6740: Yes ... it is the same for other VMs though and not specific to enabling wamr. I was working on a separate patch to do this across all git downloads consistently but can do it here now?

webui_runner/plugs/wamr_app_aot/shootout/wrapper.c
jlb6740: Ahh, yes, maybe. Unless there is something about how these preprocessor directives that won't allow this. Like some of the other comments, this will apply generally so we can test this out in a future patch that applies consistently in all wrappers.

webui_runner/Dockerfile
jlb6740: Note, I've attempted to upgrade from Bionic to Focal but it of course leads to new things breaking and more code changes. Notably wamr does not compile due to it not being able to find zlib even though it is installed:

>   Target "wamrc" links to target "ZLIB::ZLIB" but the target was not found.
>   Perhaps a find_package() call is missing for an IMPORTED target, or an
>   ALIAS target is missing?
> This warning is for project developers.  Use -Wno-dev to suppress it.

Not familiar with this error but it looks like it could be related to the cmake file included with wamr.

jlb6740:
> Looks like you want to address some of the comments in other PRs; can you create issues so we don't lose the comments?

👍 Hi @abrown Yes. I've captured https://github.com/bytecodealliance/sightglass/issues/33 https://github.com/bytecodealliance/sightglass/issues/34.

PR #31, opened 2020-07-21
record performance counter information when benchmarking

iximeow:
For the time being, Sightglass only takes advantage of this on Linux
systems. While somewhat temperamental, `perf_event_open` is a stable
interface exposed to software - Windows and MacOS don't seem to have
corresponding programmatic access to the same hardware's performance
counters. When these counters are not available, Sightglass will try
to report helpful guidance and continue, using 0 in place of reading
their measurements.

<details>
<summary>Output when performance counters are available includes cycles, instructions retired, cache accesses, and cache misses</summary>

```
lucet:                                                                                                                                                                        
  - ackermann                                                                                                                                                                  
  - base64                                                                                                                                                                    
  - ctype                                                                                                                                                                   
  - ed25519                                                                                                                                                                    
  - fib2                                                                                                                                                                      
  - gimli                                                                                                                                                                        
  - heapsort                                                                                                                                                               
  - keccak                                                                                                                                                                     
  - matrix2                                                                                                                                                                 
  - matrix                                                                                                                                                                  
  - memmove                                                                                                                                                               
  - minicsv                                                                                                                                                               
  - nestedloop2                                                                                                                                                               
  - nestedloop3                                                                                                                                                                   
  - nestedloop                                                                                                                                                                    
  - random2                                                                                                                                                                     
  - random                                                                                                                                                                   
  - ratelimit                                                                                                                                                                 
  - sieve                                                                                                                                                                      
  - strcat2                                                                                                                                                                   
  - strcat                                                                                                                                                                     
  - strchr                                                                                                                                                                 
  - strlen                                                                                                                                                                     
  - strtok                                                                                                                                                                
  - switch2                                                                                                                                                                  
  - switch                                                                                                                                                                  
  - xblabla20                                                                                                                                                                  
  - xchacha20
Test        | Implementation | Ratio               | Median       | RSD                 | Function | Percentage | CPU Cycles   | Instructions Retired | Cache Accesses | Cache Misses
------------+----------------+---------------------+--------------+---------------------+----------+------------+--------------+----------------------+----------------+-------------
ackermann   | Reference      | 1                   | 10325.5      | 241.81880162821093  |          |            | 36519.5      | 103028.5             | 4              | 1
ackermann   | lucet          | 1.5032966829622885  | 34342        | 239.35522335919416  |          |            | 125147       | 347501               | 154            | 14
base64      | Reference      | 1                   | 105913446.5  | 1.023188567250589   |          |            | 383753786.5  | 1509310168           | 31927.5        | 6944.5
base64      | lucet          | 1.7683977775335977  | 188398210.5  | 0.9009793961334853  |          |            | 676612230.5  | 2500767551.5         | 69145          | 16751.5
ctype       | Reference      | 1                   | 283220152    | 0.9119909347262768  |          |            | 1037627515.5 | 2550349334           | 202705         | 43601.5
ctype       | lucet          | 1.0742389350979755  | 303809038    | 0.9744767651821068  |          |            | 1092538464   | 4025457784           | 263335         | 56885
ed25519     | Reference      | 1                   | 574693531.5  | 1.8181208873410166  |          |            | 2108238644.5 | 6519838315.5         | 333083         | 44441.5
ed25519     | lucet          | 7.160242505759052   | 4123202872   | 0                   |          |            | 14802813123  | 50625594860          | 32284416       | 625631
fib2        | Reference      | 1                   | 615884266.5  | 0.8620538865437641  |          |            | 2269365311.5 | 5718413263.5         | 185470.5       | 29461.5
fib2        | lucet          | 2.414510266751926   | 1488897390.5 | 0.9069287494313734  |          |            | 5188418457.5 | 14307263893          | 546804         | 144329.5
gimli       | Reference      | 1                   | 25163        | 198.87449713551072  |          |            | 88442.5      | 339209.5             | 20             | 2
gimli       | lucet          | 1.6467659378558617  | 72440        | 178.40564385967664  |          |            | 268583       | 1064428              | 96             | 2
heapsort    | Reference      | 1                   | 795226447    | 0.23459338801805626 |          |            | 2923269255   | 2929795988           | 984615         | 113969
heapsort    | lucet          | 1.3196974233628704  | 1049429053   | 1.2832111331401828  |          |            | 3633201973.5 | 5911797398.5         | 2197244.5      | 468289.5
keccak      | Reference      | 1                   | 274131.5     | 153.52593072271657  |          |            | 1001392      | 3485587.5            | 192            | 28
keccak      | lucet          | 3.1586405426770154  | 1524960      | 103.17082251666044  |          |            | 4974304      | 13346394             | 2801           | 406 
matrix      | Reference      | 1                   | 203023478.5  | 4.5318786903588215  |          |            | 701582816.5  | 1726163045           | 79202          | 19567
matrix      | lucet          | 2.4160889889309822  | 482689056    | 4.7221855506203445  |          |            | 1637164331   | 6335148181           | 209291         | 85683
```
</details>

<details>
<summary>Meanwhile, when counters aren't available (hardware, kernel, `perfcnt`, permissions reasons), their values default to 0</summary>

```
lucet:                                                                                                                                                                        
Permission denied for hardware event CPUCycles. Try setting /proc/sys/kernel/perf_event_paranoid to 2 or below?                                                                
Permission denied for hardware event Instructions. Try setting /proc/sys/kernel/perf_event_paranoid to 2 or below?                                                            
Permission denied for hardware event CacheReferences. Try setting /proc/sys/kernel/perf_event_paranoid to 2 or below?                                                       
Permission denied for hardware event CacheMisses. Try setting /proc/sys/kernel/perf_event_paranoid to 2 or below?                                                              
  - ackermann                                                                                                                                                                 
  - base64                                                                                                                                                                       
  - ctype
  - ed25519                                                                                                                                                                    
  - fib2
  - gimli
  - heapsort                                                                                                                                                              
  - keccak
  - matrix2
  - matrix
  - memmove
  - minicsv                                                                                                                                                                     
  - nestedloop2                                                                                                                                                              
  - nestedloop3                                                                                                                                                               
  - nestedloop                                                                                                                                                                 
  - random2                                                                                                                                                                   
  - random                                                                                                                                                                     
  - ratelimit                                                                                                                                                              
  - sieve
  - strcat2                                                                                                                                                               
  - strcat
  - strchr                                                                                                                                                                  
  - strlen
  - strtok
  - switch2 
  - switch
  - xblabla20 
  - xchacha20 
Test        | Implementation | Ratio               | Median       | RSD                | Function | Percentage | CPU Cycles | Instructions Retired | Cache Accesses | Cache Misses
------------+----------------+---------------------+--------------+--------------------+----------+------------+------------+----------------------+----------------+-------------
ackermann   | Reference      | 1                   | 11026        | 236.06417204794056 |          |            | 0          | 0                    | 0              | 0
ackermann   | lucet          | 1.6495270251927034  | 39369        | 226.7946971340391  |          |            | 0          | 0                    | 0              | 0
base64      | Reference      | 1                   | 116059818    | 1.847703355958933  |          |            | 0          | 0                    | 0              | 0
base64      | lucet          | 1.7023970709627518  | 198949775    | 1.3444966265379914 |          |            | 0          | 0                    | 0              | 0
ctype       | Reference      | 1                   | 308272112    | 1.6243426064390174 |          |            | 0          | 0                    | 0              | 0
ctype       | lucet          | 1.0579379712332515  | 323091098    | 4.265810588977877  |          |            | 0          | 0                    | 0              | 0
ed25519     | Reference      | 1                   | 624974600    | 0.6720639448428033 |          |            | 0          | 0                    | 0              | 0
ed25519     | lucet          | 7.081425821110564   | 4420630786   | 0                  |          |            | 0          | 0                    | 0              | 0
fib2        | Reference      | 1                   | 661211571    | 0.5642755730205831 |          |            | 0          | 0                    | 0              | 0
fib2        | lucet          | 2.259444701783665   | 1493799249.5 | 1.4708579662450958 |          |            | 0          | 0                    | 0              | 0
gimli       | Reference      | 1                   | 51395        | 175.13121889153192 |          |            | 0          | 0                    | 0              | 0
gimli       | lucet          | 1.8516883077587476  | 81946        | 195.2321223509587  |          |            | 0          | 0                    | 0              | 0
```
</details>

I adjusted the json and csv output forms accordingly, as well.

There are some followups we could do here: frontend and backend stalls are available _in theory_ but they don't seem to be supported on at least my workstation. I've used these to great effect when diagnosing unexpectedly low IPC, but Sightglass might not be the place for that specific tool. I'm also very curious if folks have opinions on the "warn and continue" approach when we can't create counters for some reason.

cc @abrown and @jlb6740 who I think will also be interested in this?

edit: I think I've gotten the plumbing wrong somewhere along the way and garble up numbers. The min/max with this patch seem unreasonable, though the median didn't seem outlandish. My suspicion is that I might be measuring all iterations when benchmark functions are run multiple times, instead of measuring the function N times.


src/bench.rs
```diff
@@ -128,71 +119,172 @@ impl Runnable<()> for TestBodiesBench {
...
+    pub fn empty() -> Self {
+        Sample {
+            clock_time: Elapsed::from_ticks(0),
+            cpu_cycles: 0,
+            instructions_retired: 0,
+            cache_accesses: 0,
+            cache_misses: 0,
+        }
+    }
+}
```
abrown, line 157: I wonder if you could get away with `#[derive(Default)]` instead of this?

src/bench.rs
```diff
@@ -128,71 +119,172 @@ impl Runnable<()> for TestBodiesBench {
...
+    // disabled.
+    pub cpu_cycles: u64,
+    // Measured by performance counter. Maybe 0, in which case the counter is almost certainly
+    // disabled.
+    pub instructions_retired: u64,
+    // Measured by performance counter. Maybe 0, in which case the counter is almost certainly
+    // disabled.
+    pub cache_accesses: u64,
+    // Measured by performance counter. Maybe 0, in which case the counter is almost certainly
+    // disabled.
```
abrown, line 143: `//` -> `///` for docs?

src/perf.rs
```diff
@@ -0,0 +1,187 @@
...
+                .as_mut()
+                .and_then(|c| c.read().ok())
+                .unwrap_or(0),
+            cache_misses: self
+                .counters
+                .cache_misses
+                .as_mut()
+                .and_then(|c| c.read().ok())
+                .unwrap_or(0),
+        }
```
abrown, line 124: Is there no good way to capture the state of all of the counters at the same time? Hopefully the stuff in between each `c.read()` isn't significant but it would be nice...

jlb6740:
@iximeow Hi .. Yes, thanks, this will be a needed addition! I am thinking a feature flag would be desired for the toml. The primary reason I think is that perf adds overhead (not sure a fair number but I've read as high as 20%) and particularly when running on multiple platforms for comparison purposes (with inconsistent exposure to perf) it will be harder to compare if one platform has this overhead burden while the other doesn't.

I tried to run this the docker environment and there was an error executing the building the sightglass code. I'll dig into it, as it may be problem with the python script (sg_runner.py).

src/perf.rs
iximeow: it's.. weird. `perf_event_open` has the notion of event groups, which I _think_ are started and stopped together. The general idea is to open some perf event as the "group leader" that signals if the event group should be scheduled* or not. Then, subsequent perf events are opened with a reference to the group leader to signal to the kernel that the new events are part of the group. Unfortunately, I don't see a way to get the file descriptor for an opened event (barring wildly dangerous transmutes), which we need to pass to [`set_group`](https://docs.rs/perfcnt/0.6.0/perfcnt/linux/struct.PerfCounterBuilderLinux.html#method.set_group) to wire things up right. I'll open an upstream issue, but I think it'll take a bit of tweaking in `perfcnt` to support this.

In the mean time, my hope is that at the least the variance between these counters will be constant for a given piece of hardware, so even without that fixed I think it won't be too much noise.

<details>
<summary>*why is there a scheduler involved, you may ask? I was surprised to learn this, and now you can be too:</summary>

Some x86 performance counters are constrained to only be sampled by certain performance counters, the constraints change by microarchitecture, and Linux makes a strong effort to expose a very generic and hardware-agnostic interface to the platform's performance counters. Consequently, Linux will take the active set of requested performance counters, both in the form of human-friendly names ("CPU Cycles" or "Instructions Retired" are straightforward, "{Front,Back}end Stalls"  are more complex) and raw hardware performance counter IDs, and figure out which counters can get scheduled onto which counters to provide a fair sampling of the underlying process (or system). Because performance counters can be multiplexed between "sampling an application" and "sampling the operating system" there can be some counter-juggling on every context switch.

Additionally, and this is detail we definitely don't need to care about here, Linux exposes the idea that you may only want to sample counters in a perf event group either all together, or not at all. This makes sense when trying to count rare events that you want to measure among multiple dimensions. These impose additional custom constraints that the perf event scheduler has to respect.

The whole thing is quite a bit more complicated than I'd first expected!
</details>

iximeow:
Pardon the rebase, had to pull in some version bumps that landed in `main`. The most important change is that there is now a [`use_perf_counters`](https://github.com/bytecodealliance/sightglass/pull/31/files#diff-6d8a5e2104954a026de2bfe2b936aea4R44) field. I've also (I think) updated the UI to behave the same as it did before. I figure updating the UI and deciding how to show the new information can be a subsequent PR.

Important note: I !manually updated! the test fixture json with the format that sightglass now emits for json output. I'm not sure how or where the output other than sightglass is emitted, and wanted to raise that here. It would probably be better to update that from real reference, rather than mentally following the change through various test files.

jlb6740:
Hi @iximeow,

I took a look at this and looks like the CI error is due to perfcnt bringing in crate x86 which either needs to be built with nightly or there needs to be an update to that crate to support building with stable rust. I put a patch here: https://github.com/bytecodealliance/sightglass/pull/42 which I think should be Ok. Once this is merged this patch should pass CI as I was able to compile with nightly myself.

iximeow:
I've finally tracked down the issue (an errant `/`) and the summary information produced now looks reasonable. If anyone would like to take another look at these changes and :heavy_check_mark:, I think we're good to go.

iximeow:
the `x86` crate appears to have a new and exciting failure mode when building:
```
error: failed to run custom build command for `x86 v0.34.0`

Caused by:
  process didn't exit successfully: `/home/runner/work/sightglass/sightglass/target/debug/build/x86-98997b28b893d70c/build-script-build` (signal: 11, SIGSEGV: invalid memory reference)
```

src/perf.rs
abrown: I looked into this more today because when I used the code here in my sightglass fork I just had the nagging feeling that the repeated calls to `ioctl` before and after the section we want to measure were going to be too noisy. Maybe this is good enough for now but I ended up opening https://github.com/gz/rust-perfcnt/issues/19 to see if it is fixable in `perfcnt`.

For others who are interested, here are my notes:

If we want to use the `perfcnt` crate to measure events like CPU cycles we need to group the events or we risk polluting the results with kernel calls. `perfcnt` uses `perf_event_open` on Linux for measuring events; this means that most functionality available to the `perf` tool should be [available to users of `perfcnt`](https://github.com/gz/rust-perfcnt/blob/5f060f5253adfebbe91ceb8da0705ac6e651895a/src/linux/mod.rs#L631-L640). Notably, `perfcnt` is Linux-only (for now) because of the dependence on perf_event_open. `perf_event_open` promises the ability to measure multiple events simultaneously:
> A call to perf_event_open() creates a file descriptor that allows measuring performance information. Each file descriptor corresponds to one event that is measured; these can be grouped together to measure multiple events simultaneously (https://man7.org/linux/man-pages/man2/perf_event_open.2.html)

It does this with the `group_fd` parameter: the first event opened uses `group_fd = -1` and subsequent events use the first event's returned `fd` as their `group_fd`. When [`ioctl` is called to enable the counters](https://github.com/gz/rust-perfcnt/blob/5f060f5253adfebbe91ceb8da0705ac6e651895a/src/linux/mod.rs#L766-L772), all of the counters are measured together, which avoids measuring the overhead of enabling each event separately with `ioctl`.

As @iximeow mentions above, it is not always possible to have all the events we want scheduled on the same CPU:
> An event group is scheduled onto the CPU as a unit: it will be put onto the CPU only if all of the events in the group can be put onto the CPU.  This means that the values of the member events can be meaningfully compared—added, divided (to get ratios), and so on—with each other, since they have counted events for the same set of executed instructions. (https://man7.org/linux/man-pages/man2/perf_event_open.2.html)

But we can handpick events that we know will fit. 

perfcnt allows us to set the `group_fd` using [set_group](https://github.com/gz/rust-perfcnt/blob/5f060f5253adfebbe91ceb8da0705ac6e651895a/src/linux/mod.rs#L261-L264). But, as [someone smart on StackOverflow describes](https://stackoverflow.com/a/42092180), then we must be able to read the combined counter data appropriately: perfcnt can do this using [ReadRecord](https://github.com/gz/rust-perfcnt/blob/5f060f5253adfebbe91ceb8da0705ac6e651895a/src/linux/perf_format.rs#L198-L199) when combined with the [enable_read_format_id flag](https://docs.rs/perfcnt/0.6.0/perfcnt/linux/struct.PerfCounterBuilderLinux.html#method.enable_read_format_id). The resulting `Vec<<u64, Option<u64>)>` contained there can then be matched up to the event IDs to return a struct/map with the values associated correctly. Unfortunately, the parsing of the counter values has been limited to single events in the `ReadRecord` used in `src/linux/mod.rs` and the `ReadRecord` in `src/linux/perf_format.rs`, which looks to have the support, is not used: I opened an issue for this, https://github.com/gz/rust-perfcnt/issues/19.

iximeow: > I looked into this more today because when I used the code here in my sightglass fork I just had the nagging feeling that the repeated calls to ioctl before and after the section we want to measure were going to be too noisy. Maybe this is good enough for now but I ended up opening gz/rust-perfcnt#19 to see if it is fixable in perfcnt.

i'm inclined to agree that multiple `ioctl` will be noisier than necessary but also hope it's sufficient for us to get started. beyond that, thank you for your notes and opening the upstream issue!

abrown:
@iximeow, I think this is incorporated as a part of the `recorder` crate (i.e. the perf event parts); closing, but let me know if there are things I missed there.

PR #40, opened 2020-08-10
Switch sg-view to use NPM instead of Yarn

abrown:
This PR removes the Yarn lock file and replaces it with the NPM lock file. I have no strong preference either way but I think most non-JS contributors may be confused about the difference and I could not see how we were using Yarn here in a unique way. 

Also, this PR updates ava, a testing dependency, to get rid of some audit issues that NPM finds. This results in several warnings like `ExperimentalWarning: Conditional exports is an experimental feature. This feature could change at any time` that others seem to think are ok (e.g. https://github.com/okonet/lint-staged/issues/872).


abrown:
@jedisct1, did you have a strong preference for Yarn? Or something I'm not seeing about its use?

jedisct1:
I tend to prefer `yarn` over `npm`, but really, they do the same job, so switching to `npm` would be totally fine.

abrown:
Yeah, me too (a very weak preference); I was more aiming towards consistency with this patch since `npm` is used a couple of places but after talking to @jlb6740 I think I understand where to make the changes (i.e. CI and `webui/sg-view/Dockerfile`) so that we can use Yarn everywhere. Closing this PR...

PR #42, opened 2020-08-25
Updates the CI to use nightly when building Sightglass

jlb6740:


iximeow:
Context for Andrew and Frank: the [`perfcount`](https://github.com/bytecodealliance/sightglass/pull/31/files#diff-80398c5faae3c069e4e6aa2ed11b28c0R20) dep in #31 ends up using the `x86` crate, which itself uses `llvm_asm` (the nightly feature motivating this change)

Now, the reasons `x86` is used in `rust-perfcnt` are because the Intel SDM's description of hardware events lives there (notably no asm involved), and `rust-perfcnt` must itself embed syscalls to `perf_event_open`, since it's explicitly not bound by libc and friends.

If we're fine with `sightglass` building with nightly (I am, at the least), we can just land this and move on. We could stay on stable with some upstream changes - the syscall incantation doesn't _have_ to be inline asm (could be an external `.s` that gets built and called), but I don't have a strong enough reason to avoid nightly here.

jlb6740:
@iximeow .. I think this should be fine. To add back CI for stable we just need a feature flag that can be used to exclude features that only build with nightly.

PR #46, opened 2020-12-10
Integrate a new test runner

abrown:
This change brings in work done in a separate repository to create a new benchmark runner for sightglass. The original code is retained with a `legacy` moniker where appropriate; the new runner uses `next`. The `README.md` now contains a description of the different components present in the repository and links to documentation or implementation for each. In writing this, it clarified for me that a reorganization (in a future PR) is needed.

The new runner employs a different paradigm than the old one:
 - benchmarks are Wasm files, not shared libraries, built using Dockerfiles and committed in-tree (accompanied by their WAT pair) under `benchmarks-next`; they import `bench.start` and `bench.end` functions which are placed around the section of code to measure
 - engines (i.e. Wasm VMs) are shared libraries, like the one introduced in https://github.com/bytecodealliance/wasmtime/pull/2437, allowing the runner to control the compilation, instantiation, and execution of the Wasm benchmark
 - the building and running of the benchmarks is controlled with a CLI tool, see `cargo +nightly run -- --help` and `docs/next.md` for more information


abrown:
Obviously there are many additions and improvements to be discussed if this is merged:
 - [ ] move and rename `webui/sg-history` to `crates/result-server`
 - [ ] add the ability to run multiple iterations (e.g. `-n 10`) to `sightglass-recorder`
 - [ ] add a `sightglass-analysis` crate to aggregate and synthesize results
 - [ ] add more benchmarks! (Initially, port the existing benchmarks in shootout and polybench to the new style)
 - [ ] add a CLI command (`build-engine`?) for building engine libraries from `name@commit` strings in addition to passing paths to pre-built engines (the current approach)
 - [ ] eventually, remove/integrate duplicated parts (e.g. sightglass's test runner, webui_runner's Docker infrastructure)

fitzgen:
Nice! Excited to merge this and start checking off those TODOs and integrating with the existing web UI and all that :)

fitzgen (approved):
Rubber stamp!

Note: I did a once over of the `sightglass-next` repo and it looks like a solid foundation for adding Wasmtime benchmarking support to sightglass along the lines of RFC 4; I am assuming that this is basically the same, just with mechanical changes to move things around, so I didn't read super closely. If there is anything in particular that I should look closely at now, before this merges, let me know and I can do that.

abrown:
> just with mechanical changes to move things around

Yes, exactly. I copied over as much as I could but ended up disabling:
 - the Docker integration test (`docker.rs` uses a spawned Docker CLI process, not the Docker API which would be preferred)
 - and removing a measurement sanity check that depended on a local copy of `libwasmtime_bench_api.so`, which this repository does not yet know how to build.

Let me add a checkbox for adding tests like these back.

fitzgen:
> Obviously there are many additions and improvements to be discussed if this is merged:

To be clear: I think we should address these TODOs after we merge this PR, so that we can split them out into individual chunks of work that we can parallelize between us.

benchmarks-next/blake3-simd/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+# NPM package (https://www.npmjs.com/package/blake3), which currently uses only scalar instructions.
+RUN emcc -O3 -s STANDALONE_WASM=1 \
+    -s INITIAL_MEMORY=1048576 -s MAXIMUM_MEMORY=16777216 \
+    -s ALLOW_MEMORY_GROWTH=1 -s TOTAL_STACK=131072 \
+    -s "EXPORTED_FUNCTIONS=['_main']" \
+    -msimd128 -msse2 -msse3 -msse4.1 -msse4.2 \
+    -DBLAKE3_NO_SSE41 -DBLAKE3_NO_AVX2 -DBLAKE3_NO_AVX512 \
+    -o /benchmark.wasm \
+    blake3.c blake3_sse2.c blake3_dispatch.c \
+    blake3_portable.c benchmark.c
```
jlb6740, line 20: Maybe want to tie building this to a specific commit from BLAKE3 repository .. with similar comments for other benchmarks pulled in this way.

jlb6740:
Yes, very good work. Similar to @fitzgen, I've seen this demonstrated and the TODOs you list can all be finished next year :-). I had a small comment about fixing the commit being pulled for the benchmarks.

jlb6740 (approved):

benchmarks-next/blake3-simd/Dockerfile
fitzgen: Good call!

PR #48, opened 2020-12-12
CSV support and a few other clean ups

fitzgen:


crates/cli/src/main.rs
```diff
@@ -1,30 +1,23 @@
...
-        Err(e) => {
-            eprintln!("Error: {:?}", e);
-            process::exit(1);
-        }
-        Ok(()) => {}
-    }
+fn main() -> Result<()> {
+    pretty_env_logger::init();
+    let command = SightglassCommand::from_args();
+    command.execute()?;
```
abrown, line 16: This prints the error chain?

crates/cli/src/validate.rs
```diff
@@ -0,0 +1,26 @@
...
+        value_name = "WASMFILE",
+        parse(from_os_str)
+    )]
+    benchmark: PathBuf,
+}
+
+impl ValidateCommand {
+    pub fn execute(&self) -> Result<()> {
+        WasmBenchmark::from(&self.benchmark).is_valid()?;
+        Ok(())
```
abrown, line 24: It might be good to retain some way to print valid/invalid output for scripts that may validate multiple files. I'm fine with it being turned off by default but maybe a log::info here or in the lower levels?

crates/cli/src/main.rs
fitzgen: Yep!

```rust
use anyhow::{Context, Result};

fn bar() -> Result<()> {
    anyhow::bail!("uh oh!")
}

fn foo() -> Result<()> {
    bar().context("calling bar from foo")?;
    Ok(())
}

fn main() -> Result<()> {
    foo().context("calling foo from main")?;
    Ok(())
}
```

produces:

```
$ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.00s
     Running `target/debug/anyhow-test`
Error: calling foo from main

Caused by:
    0: calling bar from foo
    1: uh oh!
```

PR #53, opened 2020-12-15
Add support for multiple processes to the benchmark runner

fitzgen:
And a couple more little things


benchmarks-next/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+* Call `bench.start` exactly once during the execution of its `_start`
+  function. This is when the benchmark runner will start recording execution
+  time and performance counters.
+
+* Call `bench.end` exactly once during execution of its `_start` function, after
+  `bench.start` has already been called. This is when the benchmark runner will
+  stop recording execution time and performance counters.
+
+* Not use `stdout`. It may use `stderr` for logging, if necessary.
```
abrown, line 43: Why, though? If the reason is for us to be able to distinguish between sightglass output and benchmark output, we should explain that here.

benchmarks-next/blake3-scalar/rust-benchmark/src/main.rs
```diff
@@ -11,13 +11,13 @@ extern "C" {
 
 fn main() {
     let buffer = [0 as u8; 64 << 10];
-    println!("[blake3] hashing a 64KB, zero-filled buffer");
+    eprintln!("[blake3] hashing a 64KB, zero-filled buffer");
```
abrown, line 14: Where did the `benchmark.wasm` and `benchmark.wat` for this file go? Looks like they got deleted?

crates/artifact/src/docker.rs
```diff
@@ -172,7 +172,7 @@ pub struct DockerId(String);
 
 impl From<String> for DockerId {
     fn from(s: String) -> Self {
-        DockerId(s.trim().to_string())
+        DockerId(s.trim().split_whitespace().last().unwrap().to_string())
```
abrown, line 175: 👍 this is what I found was necessary to run on latest Docker (but not on my machine). This means we could likely add a CI task to do `benchmarks-next/build.sh` and then check if any benchmarks have changed to make sure the build is deterministic.

crates/cli/src/benchmark.rs
```diff
@@ -1,19 +1,128 @@
...
+    #[structopt(long = "processes", default_value = "30", value_name = "PROCESSES")]
+    processes: usize,
+
+    /// How many samples should we take from a single process?
+    #[structopt(
+        long = "samples-per-process",
+        default_value = "1",
+        value_name = "SAMPLES_PER_PROCESS"
+    )]
+    samples_per_process: usize,
```
abrown, line 36: Some thoughts about this: 
 - should the in-process and multi-process benchmark commands be the same struct? (The in-process struct could use the same `Vec<_>` parameters for engines and Wasm files)
 - if they should be separate structs, should we call this one `MultiProcessBenchmarkCommand`?
 - how are we triggering this one? The other one corresponds to the `benchmark` command... is structopt somehow smart enough to see the extra parameters and choose this struct?
 - I didn't think we needed to spell out the `short = ...` and `long = ...` attributes?

abrown (approved):

crates/cli/src/benchmark.rs
fitzgen: The `benchmark` subcommand is the multi-process runner now. The idea is that users basically won't use `in-process-benchmark`, just the regular `benchmark` subcommand will. We might even want to hide it from the help text.

I don't think we want to make the in-process benchmark commands the same. The struct definition contains some duplication, but that isn't a maintenance burden. There isn't any duplicated code, which is where things would get annoying. And I expect that as we flush out the benchmark runner, they will diverge even further.

crates/artifact/src/docker.rs
fitzgen: Yeah, getting this in CI would be great.

benchmarks-next/blake3-scalar/rust-benchmark/src/main.rs
fitzgen: Hmm... let me double check what happened here...

crates/cli/src/benchmark.rs
abrown: I see the `#[structopt(name = "benchmark")]` on `InProcessBenchmarkCommand`?

benchmarks-next/blake3-scalar/rust-benchmark/src/main.rs
fitzgen: They are still there, AFAICT?

crates/cli/src/benchmark.rs
fitzgen: Ah woops! That was a leftover and/or rebase issue. Funnily enough, it didn't seem to have an effect, as the command was still showing up as `in-process-benchmark`. Removing it now.

benchmarks-next/blake3-scalar/rust-benchmark/src/main.rs
abrown: Ok, I guess I misread the diff... must have been just a bunch of lines deleted.

crates/artifact/src/docker.rs
abrown: Let's do it as a part of #54

PR #55, opened 2020-12-15
Allow benchmarks to run multiple times with CLI flag

abrown:
@fitzgen, I will rebase this on #53 once that is merged and plumb it through the new structopt you added. Any preference on naming the new flag?
 1. `-r, --repeat`, this is what `perf stat` uses
 2. `-n, --num-iterations`, this is what I have currently in the PR
 3. `--samples` would be implied by #53 
 
I'm leaning toward option 2 and I would plumb it through as `--num-iterations-per-process` in the multi-process command.


fitzgen:
(2) sounds good to me!

fitzgen (approved):
Looks great!

PR #57, opened 2020-12-15
Update bench api

fitzgen:


crates/cli/src/benchmark.rs
```diff
@@ -149,10 +152,15 @@ pub struct InProcessBenchmarkCommand {
 impl InProcessBenchmarkCommand {
     pub fn execute(&self) -> Result<()> {
         let bytes = fs::read(&self.wasmfile).context("Attempting to read Wasm bytes")?;
+
+        let lib = libloading::Library::new(&self.engine)?;
+        let mut bench_api = unsafe { BenchApi::new(&lib)? };
```
abrown, line 157: Good call!

crates/recorder/src/benchmark.rs
```diff
@@ -94,46 +98,34 @@ impl<'a> Module<'a> {
...
+    bench_api: &mut BenchApi,
+    wasm_bytes: &[u8],
     measure_type: MeasureType,
 ) -> Result<BenchmarkMeasurements> {
-    let wasm_bytes = wasm_bytes.as_ref();
-    info!(
-        "Attempting to run benchmark of {} bytes with engine at {:?}",
-        wasm_bytes.len(),
-        engine_lib_path.as_ref()
-    );
```
abrown, line 125: Can we retain some level of logging in here? Like maybe an info in `BenchApi::new` telling us the path of the library we just loaded and maybe something here indicating that we are starting a new benchmark run (e.g. number of bytes, measure type). I find logging extremely helpful when I'm too lazy to debug.

abrown (approved):

PR #58, opened 2020-12-15
Fail CI if benchmarks change during rebuild

abrown:


.github/check-for-changes.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+if [ $(git diff --no-ext-diff --quiet --exit-code) ]; then
+    exit 0
+else
+    git log
+    git diff --name-only
+    exit 1
+fi
```
fitzgen, line 8: I think this whole thing could just be

```shell
git diff --exit-code
```

?

In which case, we might as well inline it into the CI config, rather than put it in a script.

Not sure what the `git log` is buying us, but that is going to log a lot of things. Maybe it is supposed to have `-1`? Either way, I don't think it is too useful because surely you know what the HEAD is when looking at CI.

fitzgen (approved):
r=me with the below explained and/or fixed

.github/check-for-changes.sh
abrown: Um, yeah, that is me trying to figure out what the CI is doing 😄.

abrown: Yeah, I think now that I understand this a bit better I am going to remove the script and go with `git diff --no-ext-diff --name-only --exit-code`:
 - `--no-ext-diff` just in case there's a weird diff tool setup (highly unlikely)
 - `--name-only` to print the file names (but not the bytes) so we know which files changed
 - `--exit-code` to pass/fail the build.

fitzgen:
One more thing: we might want to only do this check if anything in `benchmarks-next` has changed, so that unrelated PRs don't have to wait on this CI job.

fitzgen:
This will become more important as we add more, and larger, benchmark programs.

abrown:
> we might want to only do this check if anything in benchmarks-next has changed

That is a good point re: CI performance; my concern with this was more that the Wasm files can always be rebuilt exactly how they are currently. E.g. if we don't specify a hash in a `git clone` of some files we need to build a benchmark, a future commit to that repo will create a different Wasm file. For those of us who may need to go back in time and tweak benchmarks to see if different compilation flags, e.g., affect results, this change ensures that we still know how to build the original benchmarks.

PR #59, opened 2020-12-16
Record multiple data points per iteration and more info in each measurement

fitzgen:
@abrown, this is a WIP but I won't have time to finish before I start my holidays tomorrow. I figured it was worth posting and letting you take a look at it and give some initial feedback and/or pick it up and bring it over the finish line before you go on holiday yourself.

The idea is to remember all this extra metadata about each measurement, represent it in a way that can be translated into csv easily, and also to split up the combined perf counters data into multiple measurements.

### DIFF
--- a/crates/cli/src/benchmark.rs
+++ b/crates/cli/src/benchmark.rs
@@ -1,6 +1,7 @@
 use anyhow::{Context, Result};
 use rand::{rngs::SmallRng, Rng, SeedableRng};
 use sightglass_artifact::get_built_engine;
+use sightglass_recorder::measure::Measurements;
 use sightglass_recorder::{
     benchmark::{benchmark, BenchApi},
     measure::MeasureType,
@@ -87,11 +88,14 @@ impl BenchmarkCommand {
             }
         }
 
+        let mut csv_header = true;
+
         while !choices.is_empty() {
             let index = rng.gen_range(0, choices.len());
             let (engine, wasm, procs_left) = &mut choices[index];
 
-            let status = Command::new(&this_exe)
+            let mut command = Command::new(&this_exe);
+            command
                 .stdin(Stdio::null())
                 .stdout(Stdio::inherit())
                 .stderr(Stdio::inherit())
@@ -103,8 +107,13 @@ impl BenchmarkCommand {
                 .arg("--output-format")
                 .arg(self.output_format.to_string())
                 .arg("--num-iterations")
-                .arg(self.output_format.to_string())
-                .arg(wasm)
+                .arg(self.iterations_per_process.to_string())
+                .arg(wasm);
+            if csv_header {
+                command.arg("--csv-header");
+            }
+
+            let status = command
                 .status()
                 .context("failed to spawn benchmark process")?;
 
@@ -117,6 +126,10 @@ impl BenchmarkCommand {
             if *procs_left == 0 {
                 choices.swap_remove(index);
             }
+
+            // We only want to write the CSV header for the very first
+            // subprocess.
+            csv_header = false;
         }
 
         Ok(())
@@ -153,6 +166,12 @@ pub struct InProcessBenchmarkCommand {
     /// The format of the output data. Either 'json' or 'csv'.
     #[structopt(short = "f", long = "output-format", default_value = "json")]
     output_format: OutputFormat,
+
+    /// If we are using the CSV output format, should we write a CSV header?
+    ///
+    /// Has no effect if the output format is not CSV.
+    #[structopt(long)]
+    csv_header: bool,
 }
 
 impl InProcessBenchmarkCommand {
@@ -166,22 +185,45 @@ impl InProcessBenchmarkCommand {
         let bytes = fs::read(&self.wasmfile).context("Attempting to read Wasm bytes")?;
         log::debug!("Wasm benchmark size: {} bytes", bytes.len());
 
-        let mut measurements = Vec::with_capacity(self.iterations);
+        let arch = this_arch();
+        let engine = self.engine.to_string();
+        let wasm = self.wasmfile.display().to_string();
+        let mut measurements = Measurements::new(arch, &engine, &wasm);
+
+        let mut measure = self.measure.build();
+
         for _ in 0..self.iterations {
-            measurements.push(benchmark(&mut bench_api, &bytes, self.measure)?);
+            benchmark(&mut bench_api, &bytes, &mut measure, &mut measurements)?;
+            measurements.next_iteration();
         }
+
+        let measurements = measurements.finish();
 
         match self.output_format {
             OutputFormat::Json => {
                 println!("{}", serde_json::to_string(&measurements)?);
             }
             OutputFormat::Csv => {
-                let mut csv = csv::Writer::from_writer(io::stdout());
-                csv.serialize(&measurements)?;
+                let mut csv = csv::WriterBuilder::new()
+                    .has_headers(self.csv_header)
+                    .from_writer(io::stdout());
+                for m in &measurements {
+                    csv.serialize(m)?;
+                }
                 csv.flush()?;
             }
         }
         Ok(())
+    }
+}
+
+fn this_arch() -> &'static str {
+    if cfg!(target_arch = "x86_64") {
+        "x86_64"
+    } else if cfg!(target_arch = "aarch64") {
+        "aarch64"
+    } else {
+        unimplemented!("please add support for the current target architecture")
     }
 }
 
--- a/crates/data/src/lib.rs
+++ b/crates/data/src/lib.rs
@@ -1,0 +1,66 @@
+//! Common data definitions for sightglass.
+//!
+//! These are in one place, pulled out from the rest of the crates, so that many
+//! different crates can serialize and deserialize data by using the same
+//! definitions.
+
+#![deny(missing_docs, missing_debug_implementations)]
+
+use serde::{Deserialize, Serialize};
+use std::borrow::Cow;
+
+/// A single measurement, for example instructions retired when compiling a Wasm
+/// module.
+///
+/// This is often used with the `'static` lifetime when recording measurements,
+/// where we can use string literals for various fields. When reading data, it
+/// can be used with a non-static lifetime to avoid many small allocations.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Measurement<'a> {
+    /// The CPU architecture on which this measurement was taken, for example
+    /// "aarch64" or "x86_64".
+    pub arch: Cow<'a, str>,
+
+    /// The file path of the wasmtime benchmark API shared library used to
+    /// record this measurement.
+    pub engine: Cow<'a, str>,
+
+    /// The file path of the Wasm benchmark program.
+    pub wasm: Cow<'a, str>,
+
+    /// The id of the process within which this measurement was taken.
+    pub process: u32,
+
+    /// This measurement was the `n`th measurement of this phase taken within a
+    /// process.
+    pub iteration: u32,
+
+    /// The phase in a Wasm program's lifecycle that was measured: compilation,
+    /// instantiation, or execution.
+    pub phase: Phase,
+
+    /// The event that was measured: micro seconds of wall time, CPU cycles
+    /// executed, instructions retired, cache misses, etc.
+    pub event: Cow<'a, str>,
+
+    /// The event counts.
+    ///
+    /// The meaning and units depend on what the `event` is: it might be a count
+    /// of microseconds if the event is wall time, or it might be a count of
+    /// instructions if the event is instructions retired.
+    pub count: u64,
+}
+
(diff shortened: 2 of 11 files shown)


abrown:
We need to do something like this but I'm not a huge fan of plumbing the multiple measurements and all of the build info all the way down to into the recording code. I think it will be less of a headache in the future if we can keep each component responsible for a single thing. Here is how I was thinking of approaching this:
 - grab the machine information into something like `struct MachineInfo { name: String, arch: Arch, cpu: Cpu, ram: usize, os: Os }` (I have something like this called `BuildInfo` but we are not recording that type of build information... yet)
 - record information as we do today but then re-structure it into something like `struct BenchmarkResults<'a> { system_time: &'a SystemTime, machine_info: &'a MachineInfo, engine: &'a Engine, benchmark: &'a WasmBenchmark, process: usize, iteration: usize, phase: Phase, measurement: Measurement }`; at the end of this we would have an array of these. (I guess my thought here is not too different than what you are proposing, modulo `Cow` vs ref).
 - Use `serde`'s [struct flattening](https://serde.rs/attr-flatten.html) to create a non-hierarchical row

I think the key thing that I want to avoid from the old architecture is having to make changes throughout the entire chain when we add/remove fields, etc.

fitzgen:
This is largely where I was planning on going with this, although I wasn't thinking in terms of `serde(flatten)`, which is a good idea. When I'm caught up with my inbox and github stuff, I'll update this PR.

Thanks for the feedback!

fitzgen:
@abrown okay, there is now a `Measurements` builder instead of a `Vec<Measurement>` that lets us avoid passing tons of metadata through everything. This was the biggest missing piece in the draft PR, and I think it should address a bunch of your concerns from the last round of feedback.

I didn't end up splitting out a `MachineInfo` struct because it wasn't clear where the boundary would be (if it is just the arch then that isn't very helpful, but every other field varies at some level in the benchmarking, so it isn't clear where we should draw that line or that it would even be helpful).

Let me know what you think.

fitzgen:
Example usage and data output now:

<details>

```
$ cargo run --manifest-path ~/sightglass/crates/cli/Cargo.toml -- benchmark -e ~/wasmtime/target/release/libwasmtime_bench_api.so -f csv --num-iterations-per-process 3 --processes 3 -- ~/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm 2> /dev/null 
arch,engine,wasm,process,iteration,phase,event,count
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266007,0,Compilation,wall-cycles,361914202
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266007,0,Instantiation,wall-cycles,1062142
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266007,0,Execution,wall-cycles,14452758
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266007,1,Compilation,wall-cycles,322513785
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266007,1,Instantiation,wall-cycles,1010945
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266007,1,Execution,wall-cycles,15132543
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266007,2,Compilation,wall-cycles,320861139
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266007,2,Instantiation,wall-cycles,982282
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266007,2,Execution,wall-cycles,14545426
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266088,0,Compilation,wall-cycles,341118372
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266088,0,Instantiation,wall-cycles,1067807
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266088,0,Execution,wall-cycles,14532359
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266088,1,Compilation,wall-cycles,339891866
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266088,1,Instantiation,wall-cycles,1124546
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266088,1,Execution,wall-cycles,14404689
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266088,2,Compilation,wall-cycles,325907018
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266088,2,Instantiation,wall-cycles,1021295
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266088,2,Execution,wall-cycles,14469733
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266141,0,Compilation,wall-cycles,422688017
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266141,0,Instantiation,wall-cycles,1070057
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266141,0,Execution,wall-cycles,15561943
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266141,1,Compilation,wall-cycles,332424723
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266141,1,Instantiation,wall-cycles,1451625
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266141,1,Execution,wall-cycles,13700777
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266141,2,Compilation,wall-cycles,337276545
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266141,2,Instantiation,wall-cycles,984748
x86_64,/home/nick/wasmtime/target/release/libwasmtime_bench_api.so,/home/nick/sightglass/benchmarks-next/blake3-scalar/benchmark.wasm,266141,2,Execution,wall-cycles,16156669
```

</details>

crates/cli/src/benchmark.rs
```diff
@@ -166,25 +185,48 @@ impl InProcessBenchmarkCommand {
...
+        let engine = self.engine.to_string();
+        let wasm = self.wasmfile.display().to_string();
+        let mut measurements = Measurements::new(arch, &engine, &wasm);
+
+        let mut measure = self.measure.build();
+
         for _ in 0..self.iterations {
-            measurements.push(benchmark(&mut bench_api, &bytes, self.measure)?);
+            benchmark(&mut bench_api, &bytes, &mut measure, &mut measurements)?;
+            measurements.next_iteration();
```
abrown, line 197: Here's what I was trying to say earlier but wasn't too clear about: why not avoid passing the mutable `Measurements` all the way down into the sightglass-recorder crate? The overhead of returning a `Vec` is not important, but this would simplify the recorder logic. Instead, the "flatten to CSV rows" logic could live in the sightglass-data crate and this code would look roughly like:

```
let ms = benchmark(&mut bench_api, &bytes, &mut measure?);
measurements.insert_multiple(ms);
```

crates/cli/src/benchmark.rs
```diff
@@ -166,25 +185,48 @@ impl InProcessBenchmarkCommand {
...
         }
 
+        let measurements = measurements.finish();
+
         match self.output_format {
             OutputFormat::Json => {
                 println!("{}", serde_json::to_string(&measurements)?);
```
abrown, line 204: If the "flatten to CSV rows" logic is separated from the recording logic, then we have the option of reporting the data hierarchically with JSON and flattened with CSV. I don't know if we *should* do that, but one thing that comes to mind is that the option to report things hierarchically will make it significantly easier to interface with the current history web service (which accepts hierarchical data in a slightly different format)--less to rewrite.

crates/data/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+
+    /// The file path of the wasmtime benchmark API shared library used to
+    /// record this measurement.
+    pub engine: Cow<'a, str>,
+
+    /// The file path of the Wasm benchmark program.
+    pub wasm: Cow<'a, str>,
+
+    /// The id of the process within which this measurement was taken.
+    pub process: u32,
```
abrown, line 32: I would call this `process_id` just to make it more clear for humans reading the raw CSV.

crates/recorder/src/benchmark.rs
```diff
@@ -119,68 +124,100 @@ impl<'a, 'b> Instance<'a, 'b> {
...
+
+        let state = &mut MeasureState {
+            measure,
+            measurements,
+        };
+
+        let old_state = STATE.swap(state, Ordering::SeqCst);
+        assert!(old_state.is_null());
+
+        let result = std::panic::catch_unwind(f);
```
abrown, line 186: Learn something new about Rust every day... These are good changes! I would prefer if all of this code did not know about `Measurements`, but the concurrency handling here is great.

crates/recorder/src/measure/counters.rs
```diff
@@ -73,15 +75,17 @@ impl Measure for CounterMeasure {
...
-        })
+        measurements.reserve(4);
+        measurements.add(phase, "cpu-cycles".into(), counts[&self.cpu_cycles]);
+        measurements.add(
+            phase,
+            "instructions-retired".into(),
+            counts[&self.instructions_retired],
+        );
+        measurements.add(phase, "cache-accesses".into(), counts[&self.cache_accesses]);
+        measurements.add(phase, "cache-misses".into(), counts[&self.cache_misses]);
```
abrown, line 88: This type of thing could be moved to a `Measurement::events(&self) -> Vec<(String, u64) ` function.

abrown (commented):
My main concern is, well, the separation of concerns 😁.

crates/recorder/src/benchmark.rs
```diff
@@ -119,68 +124,100 @@ impl<'a, 'b> Instance<'a, 'b> {
...
-    let (module, compilation) = engine.compile(wasm_bytes, &mut measure);
-    info!("Compiled successfully: {:?}", compilation);
+    let module = engine.compile(wasm_bytes, measure, measurements);
+    info!("Compiled successfully");
 
     // Measure the module instantiation.
-    let (instance, instantiation) = module.instantiate(&mut measure);
-    info!("Instantiated successfully: {:?}", instantiation);
+    let instance = module.instantiate(measure, measurements);
+    info!("Instantiated successfully");
```
abrown, line 138: Another thought: I'm adding some more benchmarks and alternating between building them and running to test things out. Having the option to log the measured times is really helpful as a sanity check.

abrown (approved):

PR #60, opened 2020-12-17
Move webui/sg-history to crates/result-server

abrown:
This change moves the `sg-history` crate in with the other crates. Additionally, it adds documentation, logging, and a CI test to ensure that the Docker image is actually buildable (I wanted to also verify that the Docker image is runnable, but I can't figure out how to properly use `docker run --stop-timeout` so that the container does not run forever).

@jlb6740, this move breaks the `webui/sg_webui_*.sh` scripts; I debated on fixing them in this PR but decided to talk with you first. I think we still want something like that (i.e. some way to start everything and stop everything) but I was considering creating a combined Docker image for the Yarn app and the sg-history/sightglass-result-server service.


jlb6740:
> This change moves the `sg-history` crate in with the other crates. Additionally, it adds documentation, logging, and a CI test to ensure that the Docker image is actually buildable (I wanted to also verify that the Docker image is runnable, but I can't figure out how to properly use `docker run --stop-timeout` so that the container does not run forever).
> 
> @jlb6740, this move breaks the `webui/sg_webui_*.sh` scripts; I debated on fixing them in this PR but decided to talk with you first. I think we still want something like that (i.e. some way to start everything and stop everything) but I was considering creating a combined Docker image for the Yarn app and the sg-history/sightglass-result-server service.

@abrown Sorry I think this was submitted right after I went on break and did not notice this until recently. So based on your comment, this breaks what is currently working and used for interacting with the services? Why not keep it as a separate branch if that is the case? I haven't pulled this change to try it so I might be misunderstanding?

abrown:
@jlb6740, I think the thing to do here is to figure out if we still need the `webui/sg_webui_*.sh` scripts? I did add a Dockerfile in this PR that should be relatively equivalent to what those scripts do (maybe? Not sure about the `--volume` stuff...) and this would make it possible to keep all the Rust code under `crates`. Other benefits of this PR is that it adds documentation!

PR #61, opened 2020-12-17
Switch from perfcnt to perf-event for measuring HW counters

abrown:
Since `perfcnt` still lacks support for grouping perf event counters (see https://github.com/gz/rust-perfcnt/issues/19), this change switches to using `perf-event`, which does.

### DIFF
--- a/crates/recorder/src/measure/counters.rs
+++ b/crates/recorder/src/measure/counters.rs
@@ -1,172 +1,87 @@
-//! A mechanism to measure performance using CPU counters through `perf_event_open`.
-//!
-//! This code is lightly adapted from iximeow's work in
-//! https://github.com/bytecodealliance/sightglass/pull/31.
+//! A mechanism to measure performance using CPU counters through `perf_event_open`. For this to
+//! work (and it will only work on Linux systems currently), you may need to tweak
+//! `/proc/sys/kernel/perf_event_paranoid` by running a command such as: `sudo sysctl -w
+//! kernel.perf_event_paranoid=0`.
 use super::{Measure, Measurement};
-use perfcnt::{AbstractPerfCounter, PerfCounter};
-use precision::{Precision, Timestamp};
+use perf_event::{events::Hardware, Builder, Counter, Group};
 use serde::{Deserialize, Serialize};
 
 /// Measure CPU counters.
 pub struct CounterMeasure {
-    precision: Precision,
-    start_time: Option<Timestamp>,
-    cpu_cycles: Option<PerfCounter>,
-    instructions_retired: Option<PerfCounter>,
-    cache_accesses: Option<PerfCounter>,
-    cache_misses: Option<PerfCounter>,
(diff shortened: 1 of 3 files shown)


abrown:
@fitzgen, this is failing due to limitations on perf events in GitHub runners:
```
 ---- measure::counters::tests::sanity stdout ----
thread 'measure::counters::tests::sanity' panicked at 'Unable to create CPU_CYCLES hardware counter. Does this system actually have such a counter? If it does, your kernel may not fully support this processor.: Os { code: 1, kind: PermissionDenied, message: "Operation not permitted" }', crates/recorder/src/measure/counters.rs:21:101
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I suspect that this is similar to [this issue](https://github.community/t/getting-access-to-perf-events-on-github-hosted-ubuntu-latest/149164). Perhaps we ignore that test or remove it?

fitzgen:
> I suspect that this is similar to [this issue](https://github.community/t/getting-access-to-perf-events-on-github-hosted-ubuntu-latest/149164). Perhaps we ignore that test or remove it?

yes, that looks like the relevant issue.

I think we can check for whether [the `CI` env var](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables#default-environment-variables) is set and early exit the test if it is.

crates/recorder/src/measure/counters.rs
```diff
@@ -1,181 +1,59 @@
...
-                    cache_misses: None,
+                    cpu_cycles: Builder::new().group(&mut group).kind(Hardware::CPU_CYCLES).build().expect("Unable to create CPU_CYCLES hardware counter. Does this system actually have such a counter? If it does, your kernel may not fully support this processor."),
+                    instructions_retired: Builder::new().group(&mut group).kind(Hardware::INSTRUCTIONS).build().expect("Unable to create INSTRUCTIONS hardware counter. Does this system actually have such a counter? If it does, your kernel may not fully support this processor."),
+                    cache_accesses: Builder::new().group(&mut group).kind(Hardware::CACHE_REFERENCES).build().expect("Unable to create CACHE_REFERENCES hardware counter. Does this system actually have such a counter? If it does, your kernel may not fully support this processor."),
+                    cache_misses: Builder::new().group(&mut group).kind(Hardware::CACHE_MISSES).build().expect("Unable to create CACHE_MISSES hardware counter. Does this system actually have such a counter? If it does, your kernel may not fully support this processor."),
+                    event_group: group,
```
fitzgen, line 25: `rustfmt` won't split strings into multiple lines by default, but we can do that without introducing any extra whitespace like this:

```rust
    foo().expect(
        "long line \
         continued",
    );
```

The leading whitespace on the new line of the string is ignored.

Doing this would make these lines easier to read.

fitzgen (approved):
LGTM with tests passing as per above comment. Thanks!

PR #62, opened 2020-12-17
Run benchmarks during CI to verify benchmark-running still works

abrown:
Eventually this check should run only a smoke test version of the benchmarks to avoid over-long CI times, but the intent remains the same: we want to ensure that the tooling and benchmarks are always runnable.

### DIFF
--- a/crates/artifact/src/engine.rs
+++ b/crates/artifact/src/engine.rs
@@ -1,0 +1,239 @@
+use crate::{docker::DockerError, DockerBuildArgs, Dockerfile, GitLocation};
+use dirs::home_dir;
+use log;
+use std::{fmt, fs, path::PathBuf, str::FromStr};
+use thiserror::Error;
+
+/// Abstracts the artifact that implements the Wasm benchmarking engine. This provides the
+/// functionality for finding engine artifacts (i.e. shared libraries) and building them from
+/// Dockerfiles; eventually it should be combined with sightglass-recorder's Engine structure.
+#[derive(Debug)]
+pub enum Engine {
+    // An engine specified by its engine-ref: e.g. `wasmtime`, `wasmtime@8273ba32`,
+    // `wasmtime@some-branch@https://github.com/user/wasmtime`
+    EngineRef(EngineRef),
+    // An engine specified by its location in the file system.
+    LibraryPath(PathBuf),
+}
+impl Engine {
+    /// Calculate the path to the engine library.
+    pub fn path(&self) -> PathBuf {
+        match self {
+            Self::EngineRef(e) => Self::get_library_location(&e.to_string()),
+            Self::LibraryPath(p) => p.clone(),
+        }
+    }
+
+    /// Check if the engine library exists.
+    pub fn exists(&self) -> bool {
+        self.path().is_file()
+    }
+
+    /// If the engine is a reference, build it.
+    pub fn build(&self) -> Result<(), EngineError> {
+        log::info!("Building engine: {:?}", self);
+        if let Self::EngineRef(e) = self {
+            EngineBuilder::from(e.clone()).build()?; // TODO this is a bit backwards: we extract the engine-ref only to recreate engine
+            assert!(self.exists());
+            Ok(())
+        } else {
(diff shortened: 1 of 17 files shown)


abrown:
@fitzgen, I ended up just implementing a way to do `sightglass-cli build-engine wasmtime@[commit]@[repo]` in order to make this PR work. Roughly how this works is that sightglass finds a Dockerfile in `engines/[engine]` that it uses to build an `engine.so` file which it copies over to `~/.sightglass/[engine-ref]/engine.so`. All of this is too fragile and OS-dependent at the moment, but it works on Linux and I was hoping to iterate on this in the future. Let me know what you think.

crates/artifact/src/engine.rs
```diff
@@ -0,0 +1,239 @@
...
+/// functionality for finding engine artifacts (i.e. shared libraries) and building them from
+/// Dockerfiles; eventually it should be combined with sightglass-recorder's Engine structure.
+#[derive(Debug)]
+pub enum Engine {
+    // An engine specified by its engine-ref: e.g. `wasmtime`, `wasmtime@8273ba32`,
+    // `wasmtime@some-branch@https://github.com/user/wasmtime`
+    EngineRef(EngineRef),
+    // An engine specified by its location in the file system.
+    LibraryPath(PathBuf),
+}
```
fitzgen, line 17: I have a concern with the number of one off-types defined in this file that only exist in order to be converted into other one-off types, etc... It feels like a lot of machinery, whose logic is sort of hard to follow, when all we really need to ultimately determine is:

* do we need to build the engine via docker?
* what is the final path to the dylib?

I think we can do these things with less code, and without any one-off types, via something like this:

```rust
// Note: the `engine` parameter here would be the CLI argument, the
// returned value is a path to the built engine's dylib.
fn get_built_engine(engine: &str) -> Result<PathBuf> {
    if Path::new(engine).exists() {
        log::debug!("using already-built engine path {}", engine);
        return Ok(PathBuf::from(engine));
    }

    // Get the path to where the known engine dylib would be if it is built, or
    // else propagate an unknown engine error.
    let engine_path = known_engine_path(engine)?;

    // If no file exists at the engine path, then we have to build it.
    if !engine_path.exists() {
        log::debug!("building known engine {} at path {}", engine, engine_path.display());
        build_engine(engine)?;
        assert!(engine_path.exists());
    }

    Ok(engine_path)
}
```

fitzgen (changes requested):
Overall looks good, but I think we can tighten up the engine-building code a bit. See comments below.

engines/wasmtime/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM rust:1.48
+ARG REPOSITORY=https://github.com/bytecodealliance/wasmtime/
+ARG REVISION=main
+
+WORKDIR /usr/src
+RUN git clone ${REPOSITORY}
+WORKDIR /usr/src/wasmtime
+RUN git checkout ${REVISION} && git submodule update --init
```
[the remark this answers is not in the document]
abrown: I don't think this works; the `--recursive` will indeed pull in the submodules but the `git checkout` may be looking at an older/newer version of those submodules. We need to update the submodules after checkout--why not initialize them there too?

crates/artifact/src/engine.rs
abrown: So I attempted this tack in the latest commit. It did get rid of the weird conversions from one type to another but, because I still have to parse engine refs, I had to leave in the `EngineRef` and `WellKnownEngine` structs. I added a couple of tests to make myself feel better but I wouldn't say this latest change makes things too much clearer. Let me know if you see other ways to simplify this.

Like I noted in Zulip, one thing that complicates the issue is the there are decisions to be made at multiple levels:
 - finding an engine to use can be through either a path to a shared library or an engine-ref (e.g. wasmtime@...@...)
 - building an engine can be through either a path to a dockerfile or an engine-ref

fitzgen (approved):
The control flow is a lot clearer to me now. I think we can land this and keep iterating in tree!

PR #65, opened 2021-01-13
Allow symlink usage when building benchmarks in Docker

abrown:
This involves creating a TAR archive with any symlinks materialized and passing this to Docker via `stdin`. I believe the deduplication this allows is a weighter advantage than the potential for [symlinks being used insecurely](https://github.com/moby/moby/issues/1676).

### DIFF
--- a/crates/artifact/tests/benchmark.c
+++ b/crates/artifact/tests/benchmark.c
@@ -1,20 +1,15 @@
 #include <stdio.h>
+#include "sightglass.h"
 
 __attribute__((export_name("add")))
-__attribute__((noinline))
-int add(int a, int b) {
+__attribute__((noinline)) int
+add(int a, int b)
+{
     return a + b;
 }
 
-__attribute__((import_module("bench")))
-__attribute__((import_name("start")))
-void bench_start();
-
-__attribute__((import_module("bench")))
-__attribute__((import_name("end")))
-void bench_end();
-
-int main() {
+int main()
+{
     bench_start();
     int c = add(40, 2);
     bench_end();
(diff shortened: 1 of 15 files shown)


benchmarks-next/Dockerfile.default
```diff
@@ -0,0 +1,7 @@
+FROM emscripten/emsdk:2.0.8
```
fitzgen, line 1: Maybe name this file `EmscriptenDockerfile.default` so we can eventually have similar deduplication for rust containers?

crates/artifact/tests/benchmark.c
```diff
@@ -1,20 +1,15 @@
...
-__attribute__((import_module("bench")))
-__attribute__((import_name("start")))
-void bench_start();
-
-__attribute__((import_module("bench")))
-__attribute__((import_name("end")))
-void bench_end();
-
-int main() {
+int main()
```
fitzgen, line 11: Let's either keep the return type on the same line as the function name or not. Having one function that does that and another that doesn't is inconsistent and makes the code harder to read.

fitzgen (approved):
:+1:

crates/artifact/tests/benchmark.c
abrown: Bleh, yeah. I switched IDEs and the new one has weird defaults. (This also reminds me that I should add a Rust lint check in the CI... not sure about a C lint check).

benchmarks-next/Dockerfile.default
abrown: I went with `Dockerfile.emscripten` since I've seen/used that name format before.

PR #68, opened 2021-01-15
Convert shootout benchmarks over to benchmarks-next

abrown:
This involves (as codified in the `convert.sh` script):
 - creating a new directory for each micro-benchmark and linking to the default Dockerfile and the `sightglass-next.h` header
 - copying over the benchmark from `benchmarks/shootout`
 - adding `bench_start()` and `bench_end()` in (hopefully) the right places
 - changing `[benchmark]_body()` to `main()`
 - sometimes adding `printf()` statements, for benchmarks that have an output


abrown:
One thing that @fitzgen has mentioned in the past is that we will want to send in the input to the benchmark in some cases to avoid compiler optimizations based on the knowledge of a variable. This PR does not make those changes--it's more of a mechanical translation from the old style to the new style.

Since it's on my mind, we could follow a scheme similar to [what is being proposed for WASI testing](https://github.com/WebAssembly/WASI/pull/373/files#diff-09674199b1af4ad089fcfda544f7434c79c0eacc641b39500a70a08f96ae36bbR8-R16); i.e. files for stdin, stdout, args, etc.

abrown:
Other things that should happen eventually:
 - we should ensure these benchmarks have been through wasm-opt or the like
 - I didn't include some benchmark variants, e.g. matrix2, nestedloop2, etc.; I'm hoping we can just decide which variant is best and use that

fitzgen:
Rebased and ported to WASI in https://github.com/bytecodealliance/sightglass/pull/81

PR #69, opened 2021-01-15
Refactor to avoid threading Measurements

abrown:
This commit, which builds on the work of #59, is an incomplete refactoring: I didn't exactly know how to set up the lifetimes of `static_measure::MeasureState` and use this correctly with the `AtomicPtr` changes (so I don't expect this to compile). I left some `todo!`s in there pending your comments. If you explain what should be done there, I don't mind eventually merging this, but I would like to clean up and rename some things; this PR is mainly to explain what I meant.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1353,6 +1353,13 @@ dependencies = [
 ]
 
 [[package]]
+name = "sightglass-data"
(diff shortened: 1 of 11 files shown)


abrown:
Not needed.

PR #72, opened 2021-01-26
Fix: perf counters must be reset on Measure::start and disabled on Measure::end

abrown:
@fitzgen, the `disable` call was removed in the last set of commits; `reset` was not present but should be if we are going to reuse `Measure`s within a process.


fitzgen (approved):
Whoops! Sorry about that!

Issue #70, opened 2021-01-21
sightglass next: Support for multiple input workloads


fitzgen:
After chatting with @abrown we came up with the following extension to the benchmark program protocol to support multiple input workloads (eg a "default" workload and a "small" workload that is used for quickly testing an idea):

* the runner has a `--small-workload` flag
* this sets a `WASM_BENCH_USE_SMALL_WORKLOAD=1` environment variable for benchmark programs
* the directory where the wasm benchmark program lives is preopened for its execution
* the wasm benchmark program may read that env var and use that to decide which input files to read and process as its workload

Example:

```
fn main() {
    let workload = if std::env::var("WASM_BENCH_USE_SMALL_WORKLOAD").is_ok() {
        read_small()
    } else {
        read_default()
    };
    bench_start();
    process(workload);
    bench_end();
}
```

Things to note:

* benchmarks can ignore the env var if they have only a single workload and it Just Works (eg micro benchmarks)
* easier to implement in runner than a stdin protocol, work is pushed to benchmark programs themselves, and it is easier for them then the runner
* flexible for situations where a workload is made up of multiple files, where as stdin would require some sort of JSON wrapping of them or something
* easy to add new env vars in the future for new extension points, if needed

abrown:
@fitzgen, isn't this complete?

fitzgen:
Yes, although not all benchmarks take advantage of it (which is fine).

PR #73, opened 2021-01-26
sightglass-next: Add support for using small workloads

fitzgen:
Pass the `--small-workloads` flag to set the `WASM_BENCH_USE_SMALL_WORKLOAD` env var so that benchmark programs can use smaller, faster-to-execute inputs, if they have them available. I added small workloads for the blake3 programs.


fitzgen:
Fixes #70

fitzgen:
Hm looks like building the ackermann program doesn't use WASI's stat, but something else (I'm assuming it is emscripten's internal syscall).

I can split that last commit out into a new PR so that we don't hold up the rest of this stuff.

fitzgen:
Getting the C benchmarks working requires using wasi-sdk in docker instead of emscripten.

fitzgen:
I have the ackermann benchmark working with wasi-sdk, but the blake3-simd benchmark is proving difficult.

abrown (approved):

abrown:
ci: blake-scalar is not finding `default.input`?

fitzgen:
Hm, interesting... it runs fine locally :thinking:

fitzgen:
I think the issue was building the default wasmtime engine via docker, which had previously been happening in child processes and was using a relative path to the dockerfile, and now that the child processes are in a different working directory, that relative path broke. The fix is to build engines up front, before we spawn child processes.

fitzgen:
Gah, no, that was a different bug. Still hunting down the CI bug...

fitzgen:
Okay, I've got this green in CI now, but it needs https://github.com/bytecodealliance/wasmtime/pull/2612 to merge before I can remove the "TEMP" commit.

fitzgen:
Now that the wasmtime PR merged this should be good to merge.

Issue #71, opened 2021-01-22
sightglass next: if no wasm files are passed to the `benchmark` subcommand, default to all benchmark programs


fitzgen:
If none are passed, then we should default to `./benchmarks-next/*/benchmark.wasm` to run every available benchmark.

PR #76, opened 2021-01-28
Add initial 'sightglass-cli analyze' command

abrown:
This adds functionality to create aggregated statistics from serialized measurements. In this early incarnation, it is an additional `sightglass-cli` command that parses STDIN and aggregates min, max, mean, and mean deviation after grouping the measurements by phase and event type.

> Note: this should be merged after #75 as it depends on those changes.

### DIFF
--- a/crates/analysis/src/lib.rs
+++ b/crates/analysis/src/lib.rs
@@ -1,0 +1,167 @@
+use sightglass_data::{Measurement, Phase, Summary};
+use std::{borrow::Cow, collections::BTreeSet};
+
+/// Summarize measurements grouped by: architecture, engine, benchmark file, phase and event.
+pub fn summarize<'a>(measurements: &[Measurement<'a>]) -> Vec<Summary<'a>> {
+    let mut summaries = Vec::new();
+    for k in keys(&measurements) {
+        let grouped_counts: Vec<_> = measurements
+            .iter()
+            .filter(|m| k.matches(m))
+            .map(|m| m.count)
+            .collect();
+        summaries.push(Summary {
+            arch: k.arch,
+            engine: k.engine,
+            wasm: k.wasm,
+            phase: k.phase,
+            event: k.event,
+            min: grouped_counts
+                .iter()
+                .cloned()
+                .min()
+                .expect("at least one element"),
+            max: grouped_counts
+                .iter()
+                .cloned()
+                .max()
+                .expect("at least one element"),
+            mean: mean(&grouped_counts),
+            mean_deviation: mean_deviation(&grouped_counts),
+        })
+    }
+    summaries
+}
+
+/// Extract the keys (i.e. `phase` and `event`) for the groups of measurements to aggregate.
+fn keys<'a>(measurements: &[Measurement<'a>]) -> Vec<Key<'a>> {
+    let set: BTreeSet<_> = measurements
+        .iter()
+        .cloned()
+        .map(|m| Key {
+            arch: m.arch,
+            engine: m.engine,
+            wasm: m.wasm,
+            phase: m.phase,
+            event: m.event,
+        })
+        .collect();
+    set.into_iter().collect()
+}
+#[derive(PartialOrd, Ord, PartialEq, Eq, Hash)]
+struct Key<'a> {
+    arch: Cow<'a, str>,
+    engine: Cow<'a, str>,
+    wasm: Cow<'a, str>,
+    phase: Phase,
+    event: Cow<'a, str>,
+}
+impl Key<'_> {
+    fn matches(&self, m: &Measurement) -> bool {
+        self.arch == m.arch
+            && self.engine == m.engine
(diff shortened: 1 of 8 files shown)


fitzgen (commented):
The code itself looks fine, but I think it is worth exploring what we want the final UX to look like for analysis and how this code fits into that vision, before we go too much further.

Both our two main use cases (which are very similar to each other) involve comparing two versions of wasmtime:

* CI comparing main vs a feature branch
* Local devs comparing main vs a feature branch

Both of these will record data something like

```
sightglass benchmark -e wasmtime@main -e wasmtime@feature-branch
```

which will produce a big blog of data where measurements from `main` and from `feature-branch` are intermingled, as well as results from a bunch of different benchmark programs (once we implement #71).

So, right off the bat, the assumption that the data is all from the same "execution context" doesn't hold.

I think that instead of checking that all the data is in the same execution context and then erroring out if it isn't, we should automatically group the measures by `(arch, engine, benchmark, phase)` and report the summary for each. Would you be up for this?

Later we can add ANOVA and effect sizes for each group on top of this as well.

crates/analysis/src/lib.rs
```diff
@@ -0,0 +1,127 @@
+use sightglass_data::{Measurement, Phase, Summary};
+use std::{borrow::Cow, collections::HashSet};
+
+/// Summarize measurements grouped by: architecture, engine, benchmark file, phase and event.
+pub fn aggregate<'a>(measurements: &[Measurement<'a>]) -> Vec<Summary<'a>> {
+    let mut summaries = Vec::new();
+    for (arch, engine, wasm, phase, event) in keys(&measurements) {
+        let grouped_counts: Vec<_> = measurements
+            .iter()
+            .filter(|m| m.phase == phase && m.event == event)
```
fitzgen, line 10: I think this needs to filter `m.arch == arch` and similar for `engine` and `wasm` as well?

crates/analysis/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+/// Extract the keys (i.e. `phase` and `event`) for the groups of measurements to aggregate.
+fn keys<'a>(measurements: &[Measurement<'a>]) -> Vec<Key<'a>> {
+    let set: HashSet<_> = measurements
+        .iter()
+        .cloned()
+        .map(|m| (m.arch, m.engine, m.wasm, m.phase, m.event))
+        .collect();
+    set.into_iter().collect()
+}
+type Key<'a> = (
```
fitzgen, line 45: It probably makes sense to just define a proper `struct` for this, since we have to explain what "key" means in the doc comment but then also its definition is too big to do inline, so we have a type alias. Instead, if we just define a proper `struct`, we can get rid of the mini-explanation of what a key is in the doc comment and we have a single place where we need to explain what it is.

crates/analysis/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+            mean: mean(&grouped_counts),
+            deviation: deviation(&grouped_counts),
+        })
+    }
+    summaries
+}
+
+/// Extract the keys (i.e. `phase` and `event`) for the groups of measurements to aggregate.
+fn keys<'a>(measurements: &[Measurement<'a>]) -> Vec<Key<'a>> {
+    let set: HashSet<_> = measurements
```
fitzgen, line 38: This should probably be a `BTreeSet` instead of a `HashSet` so that we have deterministic ordering for tests and all that.

crates/data/src/lib.rs
```diff
@@ -67,3 +67,38 @@ pub enum Phase {
...
+    pub min: u64,
+
+    /// The maximum value of the `count` field.
+    pub max: u64,
+
+    /// The arithmetic mean of the `count` field.
+    pub mean: f64,
+
+    /// The mean deviation (note: not standard deviation) of the `count` field.
+    pub deviation: f64,
```
fitzgen, line 103: Should this be named `mean_deviation` instead of plain `deviation`? Will be more consistent/clear if we ever want to add `std_deviation` as well.

crates/analysis/src/lib.rs
```diff
@@ -0,0 +1,127 @@
...
+    Cow<'a, str>,
+);
+
+/// Calculate the arithmetic mean of a slice of numbers.
+fn mean(numbers: &[u64]) -> f64 {
+    numbers.iter().sum::<u64>() as f64 / numbers.len() as f64
+}
+
+/// Calculate the mean deviation (note: not standard deviation) of a slice of numbers.
+fn deviation(numbers: &[u64]) -> f64 {
```
fitzgen, line 59: ditto

fitzgen (approved):
LGTM! Feel free to merge once comments below are addressed

crates/analysis/src/lib.rs
abrown: Oh, right; let me update the test to catch that.

PR #77, opened 2021-01-28
Document overhead of the bench.start/bench.end API

abrown:
This relies on the functionality of #75 and #76 to create this documentation, but since it is only documentation, it is not dependent Git-wise on those changes.


fitzgen (approved):
LGTM :+1:

PR #84, opened 2021-02-18
Add remaining shootout benchmarks

abrown:
Add the remaining shootout benchmarks (no external input yet).


fitzgen (approved):

abrown:
Questions for @jedisct1:
 - which shootout benchmarks actually need LICENSE files? (Which ones did you not create?)
 - if you have any notes about these benchmarks I would be happy to organize them into some README files; e.g. why this benchmark is a representative workload, what features are being measured by this other benchmark, etc.
 - I didn't include the `str*` benchmarks because I don't understand their purpose exactly and I felt like I had seen them fail on various engines/platforms in runs @jlb6740 had done--do we need them?
 - I also didn't include the duplicate version of some benchmarks (e.g. `matrix2.c`, `random2.c`, `switch2.c`) also because it was hard to discern why the various versions were needed--do we need them?

jedisct1:
`ackerman`, `fib2`, `matrix` and `sieve`  need a license file.

The `str*` benchmarks were made to measure the impact of a feature in Lucet where calls to specific functions could be replaced by pre-optimized builtins. With bulk memory operations, this is probably not relevant any more.

The difference between `matrix` and `matrix2` is that one of them accesses arrays whose size is known at compile time, while the other one uses dynamic arrays. That was to exercise bound checking elision.

`switch2` is indeed redundant.

PR #85, opened 2021-02-22
Add a sightglass-api crate

abrown:
Provides a helper crate for exposing `bench::start()` and `bench::end()` to Rust code.

Some things that I was debating on changing:
 - this crate is separated from `include/sightglass-next.h` but they really need to be kept in sync--I can't really think of a good way to do this now, though
 - we could probably guarantee from the Sightglass end that `start` and `end` are safe but this API still requires the use of `unsafe`--thoughts?
 - we might think about adding a `black_box` function to this API to match `include/sightglass-next.h`.


benchmarks-next/blake3-scalar/rust-benchmark/Cargo.toml
```diff
@@ -5,5 +5,6 @@ edition = "2018"
 
 [dependencies]
 blake3 = "0.3.7"
+sightglass-api = { path = "../../../crates/api" }
```
abrown, line 8: I think we would probably have to publish the sightglass-api crate, not use a local path, since these benchmarks are built inside Docker.

fitzgen (commented):
Is this worth the hassle? It saves a small handful of lines of code that would otherwise be duplicated, but comes with the headache of needing to publish this crates to crates.io so it is visible in docker.

To make it the API safe, which we would want to do if we pursue this further, we'd need to wrap these functions like:

```rust
mod ffi {
    #[link(wasm_import_module = "bench")]
    extern "C" {
        pub fn start();
        pub fn end();
    }
}

pub fn start() {
    unsafe {
        ffi::start();
    }
}

pub fn end() {
    unsafe {
        ffi::end();
    }
}
```

abrown:
Yeah, the publishing part is a bit of a pain... how painful, though? Having this type of thing helps me not forget what code I have to copy and paste around when I try out various things with the `sightglass-cli` (which works great from `cargo install`, BTW).

fitzgen:
Yeah, it's probably fine. r=me with the safety fix I suggested above and adding the wasmtime publishers team as another owner on crates.io

fitzgen (approved):

PR #86, opened 2021-02-24
Add a --working-dir flag for configuring the benchmark's working directory

abrown:


fitzgen (commented):
Using `"."` as a default will break the typical use case of

```
$ sightglass benchmark -- path/to/benchmark.wasm
```

where the default working dir should be `path/to` in this case. If we want to support this, then I think we want `working_dir: Option<PathBuf>` and set it to `self.wasmfile.parent()` if not supplied.

Why do you find yourself needing to customize the working dir in the first place?

abrown:
The current logic enforces that the working directory must be the same as the benchmark file's directory. I attempted to run a benchmark using a different directory, e.g. `cd context-dir; sightglass-cli benchmark -- ../path/to/benchmark.wasm` and kept on getting errors indicating that files could not be opened. This was confusing, since I had changed to the directory with the files I wanted to use!

I had to read through the code to figure out this "the context dir must be the same as the benchmark parent" assumption. I think that assumption is a bit limiting, so I added `--working-dir` to make this more visible (and configurable) to the user.

I see what you mean about `.`--using the Wasm file parent as the default is fine. I'll change that.

fitzgen:
> The current logic enforces that the working directory must be the same as the benchmark file's directory. I attempted to run a benchmark using a different directory, e.g. `cd context-dir; sightglass-cli benchmark -- ../path/to/benchmark.wasm` and kept on getting errors indicating that files could not be opened. This was confusing, since I had changed to the directory with the files I wanted to use!

FWIW, benchmarks that *don't* use their current directory will break our future bench-everything workflow (https://github.com/bytecodealliance/sightglass/issues/71) since we won't be able to specify working dirs for just one out of N benchmarks.

Why do you want to have the extra files/directories not within the `benchmarks-next/foobar/*` directory or a subdirectory of that? It seems out of place to have the inputs be at `../blah` relative to the benchmark file itself.

abrown:
This is to make `sightglass-cli` useful for running workloads outside of this repository; in those cases, the Wasm file and the working directory may not always be the same. This seemed like a way to make that easier.

Within the sightglass repository itself, I agree we should just keep all of the inputs in the same directory as the Wasm file, like we are currently doing.

fitzgen:
Ah okay, gotcha.

fitzgen (approved):

crates/cli/src/benchmark.rs
```diff
@@ -206,11 +212,17 @@ impl InProcessBenchmarkCommand {
...
         let mut bench_api = unsafe { BenchApi::new(&lib)? };
 
-        let working_dir = if let Some(dir) = self.wasmfile.parent() {
-            dir.into()
+        // Use the provided --working-dir, otherwise find the Wasm file's parent directory.
+        let working_dir = if let Some(dir) = self.working_dir.clone() {
+            dir
         } else {
-            std::env::current_dir().context("failed to get the current working directory")?
+            if let Some(dir) = self.wasmfile.parent() {
```
fitzgen, line 219: This can be `} else if let Some(dir) = ... {`

PR #87, opened 2021-02-24
Add additional 'nanoseconds' field to 'wall-cycles' measure

abrown:
Having both `cycles` and `nanoseconds` (calculated using the processor speed measured by the `precision` crate) is quite useful.


fitzgen (approved):
:+1:

PR #88, opened 2021-02-25
Add a markdown rendering benchmark program

fitzgen:


abrown (approved):
Makes sense to me! (Good catch with blake3).

PR #100, opened 2021-03-15
Unify CLI options, removing 'in-process-benchmark'

abrown:
Prior to this change, the CLI exposed an `in-process-benchmark` command as well as a `benchmark` command (that spawns processes using `in-process-benchmark`). To avoid confusion, `in-process-benchmark` is removed and the functionality is retained by calling `benchmark --processes 1`.


fitzgen (approved):
LGTM!

PR #104, opened 2021-03-19
Avoid re-downloading wasi-sdk when building benchmarks

### LINKED ISSUE #101: sightglass-next build process downloads WASI SDK many times
While watching the output of `benchmarks-next/build.sh` scroll by, I noticed that it seems that every individual benchmark build is reconstructing a Docker image from scratch, including a download of the 48MB WASI SDK tarball as well as a number of Ubuntu packages. This is somewhat wasteful (bandwidth is not free) and also means the build is slower than it could be.

Could we consider building a base Docker image once and then using it to build each benchmark in turn?

abrown:
Fixes #101 by using [multi-stage builds](https://docs.docker.com/develop/develop-images/multistage-build). Instead of requiring us to publish Docker images for this type of thing (and keep them up-to-date), this should retrieve wasi-sdk only the first time and use cached images for subsequent runs.

### DIFF
--- a/benchmarks-next/Dockerfile.wasi-sdk
+++ b/benchmarks-next/Dockerfile.wasi-sdk
@@ -1,13 +1,16 @@
-FROM ubuntu:18.04
 
+# This two-phase Dockerfile allows us to avoid re-downloading APT packages and wasi-sdk with every
+# build. First, retrieve wasi-sdk:
+FROM ubuntu:18.04 AS builder
 WORKDIR /
-
 RUN apt update && apt install -y wget
-
 # Download and extract wasi-sdk.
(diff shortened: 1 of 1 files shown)


abrown:
A good indicator of whether this is successful is if it reduces the time take to rebuild the benchmarks from [the current 14-24m range](https://github.com/bytecodealliance/sightglass/actions/workflows/benchmarks.yml).

abrown:
And yes, looks like the rebuild task is only taking 9m and using cached images.

cfallin (approved):
Nice, I didn't know about multi-stage Dockerfiles. Thanks!

PR #105, opened 2021-03-23
Allow specifying file outputs + enhance benchmarks-next/run.sh

bnjbvr:
I was trying to use sightglass for the first time, to measure a difference in terms of compile time for a specific change I'm making. I've added enough code so one can output the results to specific file paths (instead of stdio). I've also tweaked the main script so it'll store the results to a file, then these results can be analyzed using raw `jq` queries. This is a bit hacky, but it helps immediately answering questions like "what's the compile time for each benchmark". See the updates to the documentation for examples and new use cases.

### DIFF
--- a/crates/cli/src/analyze.rs
+++ b/crates/cli/src/analyze.rs
@@ -1,7 +1,10 @@
 use anyhow::Result;
 use sightglass_analysis::summarize;
 use sightglass_data::{Format, Measurement};
-use std::io;
+use std::{
+    fs::File,
+    io::{self, BufReader, Read},
+};
 use structopt::StructOpt;
 
 /// Analyze benchmark output; accepts raw benchmark results in `stdin` (i.e., from `sightglass-cli
@@ -13,6 +16,10 @@ pub struct AnalyzeCommand {
     #[structopt(short = "i", long = "input-format", default_value = "json")]
     input_format: Format,
 
+    /// Path to the file that will be read from, or none to indicate stdin (default).
+    #[structopt(short = "f")]
+    input_file: Option<String>,
+
     /// The format of the output data. Either 'json' or 'csv'.
     #[structopt(short = "o", long = "output-format", default_value = "json")]
     output_format: Format,
@@ -20,7 +27,12 @@ pub struct AnalyzeCommand {
 
 impl AnalyzeCommand {
     pub fn execute(&self) -> Result<()> {
-        let measurements: Vec<Measurement> = self.input_format.read(io::stdin())?;
+        let file: Box<dyn Read> = if let Some(file) = self.input_file.as_ref() {
+            Box::new(BufReader::new(File::open(file)?))
+        } else {
+            Box::new(io::stdin())
+        };
+        let measurements: Vec<Measurement> = self.input_format.read(file)?;
         let summaries = summarize(&measurements);
         self.output_format.write(&summaries, io::stdout())
     }
--- a/crates/recorder/src/benchmark.rs
+++ b/crates/recorder/src/benchmark.rs
@@ -135,6 +135,8 @@ pub fn benchmark(
     bench_api: &mut BenchApi,
     working_dir: &Path,
     wasm_bytes: &[u8],
+    // Optionally stop after the given phase, rather than running all phases.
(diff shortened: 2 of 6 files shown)


benchmarks-next/run.sh
```diff
@@ -5,11 +5,33 @@ PROJECT_DIR=$(dirname "$0" | xargs dirname)
 SIGHTGLASS="cargo +nightly run --bin sightglass-cli --"
 export RUST_LOG=debug
 
-$SIGHTGLASS build-engine wasmtime
+ENGINE=${ENGINE:-wasmtime}
```
fitzgen, line 8: `run.sh` is a temporary placeholder, and any improvements to it, should instead be made to the benchmark runner CLI itself.

crates/cli/src/analyze.rs
```diff
@@ -13,14 +16,23 @@ pub struct AnalyzeCommand {
     #[structopt(short = "i", long = "input-format", default_value = "json")]
     input_format: Format,
 
+    /// Path to the file that will be read from, or '-' to indicate stdin (default).
+    #[structopt(short = "f", default_value = "-")]
+    input_file: String,
```
fitzgen, line 21: Can we do `input_file: Option<String>` and have `None` be `stdin`? This way we don't have to invent syntax for stdin (which I never liked about `clif-util`, it meant you always had to remember to check for `"-"` manually rather than letting the type system guide you) and we can support files named `-` (even if anyone actually making such a file is a terrible idea).

crates/cli/src/analyze.rs
```diff
@@ -13,14 +16,23 @@ pub struct AnalyzeCommand {
...
     output_format: Format,
 }
 
 impl AnalyzeCommand {
     pub fn execute(&self) -> Result<()> {
-        let measurements: Vec<Measurement> = self.input_format.read(io::stdin())?;
+        let file: Box<dyn Read> = if self.input_file == "-" {
+            Box::new(io::stdin())
+        } else {
+            Box::new(File::open(&self.input_file)?)
```
fitzgen, line 33: Any reason not to wrap these in a `BufReader`?

fitzgen (commented):
Thanks Benjamin!

A couple comments below.

crates/cli/src/analyze.rs
fitzgen: (Same for the other subcommand and `output_file`)

benchmarks-next/run.sh
bnjbvr: Well, this was useful to run a bunch of wasm modules, but I'll remove this commit then.

crates/recorder/src/benchmark.rs
```diff
@@ -128,13 +128,19 @@ impl<'a, 'b> Instance<'a, 'b> {
     }
 }
 
+pub struct BenchParameters {
+    pub must_instantiate: bool,
+    pub must_execute: bool,
+}
```
fitzgen, line 134: Having `must_instantiate == false && must_execute == true` is a bit nonsensical; it would be nice if the type couldn't represent invalid states. Therefore, I think it would be nicer if we had something like the following:

```rust
use sightglass_data::Phase;

pub fn benchmark(
    // existing params...

    // Optionally stop after the given phase, rather than running all phases.
    stop_after: Option<Phase>,
) -> Result<()> {
    // ...
}
```

engines/wasmtime/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM rust:1.48
+FROM rust:1.51
```
fitzgen, line 1: FWIW, this won't be available until https://github.com/docker-library/official-images/pull/9856 merges.

fitzgen (approved):
r=me with a nitpick below

Thanks @bnjbvr!

crates/recorder/src/benchmark.rs
bnjbvr: Good call; even allowing the combination of CLI flags `--execute true --instantiate false` was a bit misleading, so I've changed it so that one directly pass the `stop_after` Phase option to the command line.

PR #111, opened 2021-04-02
Print CSV headers

### LINKED ISSUE #107: CSV headers no longer showing up
At some point in time CSV headers stopped being printed. We should fix that.

abrown:
This change adds a field in `Format::Csv` to indicate when headers should be printed. Since headers should only be printed once in a sequence of benchmark runs, the field is internally mutable--the headers are printed or read for the first record only and never again. This seems like an internal detail so no header flag is exposed in the CLI (presumably it should be easy for users to discard the first line if headers are not desired). Fixes #107.


fitzgen (approved):
Nice, thanks!

PR #121, opened 2021-05-13
Update the checked-in version of pulldown-cmark

abrown:
@fitzgen, this benchmark has been failing in the "check that the benchmarks rebuild derministically" test. Any ideas what is changing in the build that could cause the 3-byte diff?

(Also, let me know if you want to re-build this with Rust v1.51 instead of Rust v1.45 currently specified in the Dockerfile?)


fitzgen:
> @fitzgen, this benchmark has been failing in the "check that the benchmarks rebuild derministically" test. Any ideas what is changing in the build that could cause the 3-byte diff?

I don't know... :-/

Maybe it uses some kind of build.rs somewhere in the dep tree that isn't deterministic.

PR #124, opened 2021-05-21
Add support for effect sizes and confidence intervals

### LINKED ISSUE #89: sightglass-next: add statistical significance test
Implement (or re-use) some Rust code that will be able to test whether the performance difference between two runs (i.e., the aggregate runs of two different engines) is statistically significant or not. Probably involves some additions to the `sightglass-analysis` crate and potentially some refactoring.

### LINKED ISSUE #90: sightglass-next: add effect size confidence intervals
Implement a statistically sound way (we discussed doing this in Rust) to determine the confidence intervals of the speedup/slowdown observed between two engines.

fitzgen:
Fixes #90.

Fixes #89 (if the baseline is within the effect size's CI then it isn't statistically significant).


crates/data/src/lib.rs
```diff
@@ -117,3 +117,75 @@ pub struct Summary<'a> {
...
+
+    /// The event that was measured: micro seconds of wall time, CPU cycles
+    /// executed, instructions retired, cache misses, etc.
+    pub event: Cow<'a, str>,
+
+    /// The first engine variant.
+    pub a: EffectSizeVariant<'a>,
+
+    /// The second engine variant.
+    pub b: EffectSizeVariant<'a>,
```
abrown, line 148: Not sure if the multi-level structuring here will show up in the CSV output. Do you know? I feel like sometime ago we switched to single-level output structs for this reason. (We could add a test in `crates/data/tests` to check this?).

crates/data/src/lib.rs
```diff
@@ -117,3 +117,75 @@ pub struct Summary<'a> {
...
+    /// The file path of the wasmtime benchmark API shared library used to
+    /// record this measurement.
+    pub engine: Cow<'a, str>,
+
+    /// The arithmetic mean of the `count` field.
+    pub mean: f64,
+}
+
+impl EffectSize<'_> {
+    /// Return `b`'s speed up over `a` and the speed up's confidence interval.
```
abrown, line 176:
```suggestion
    /// Return `b`'s speedup over `a` and the speedup's confidence interval.
```

crates/data/src/lib.rs
```diff
@@ -117,3 +117,75 @@ pub struct Summary<'a> {
     /// The mean deviation (note: not standard deviation) of the `count` field.
     pub mean_deviation: f64,
 }
+
+/// The effect size (and confidence interval) of between two different engines
```
abrown, line 121:
```suggestion
/// The effect size (and confidence interval) between two different engines
```
Maybe this is what is meant here?

abrown (approved):

crates/data/src/lib.rs
fitzgen: Good call, I forgot that nested structs aren't automatically flattened with the `csv` crate. Pushed a fix.

PR #127, opened 2021-05-25
Update for new benchmarking API

fitzgen:
This means that we can now pass in file paths where stdout/stderr should be
logged to, an (optional) file path that can be read and passed to Wasm execution
as its stdin, and we can instantiate the module multiple times per
compilation (although this feature is not hooked up to the CLI yet).

The management of starting and ending timers is also simplified with the new
benchmark API, and so we have no more need for the `static_measure` module and
its trickery.

See also https://github.com/bytecodealliance/wasmtime/pull/2937


fitzgen:
Subsequent PRs will add the checking of expected stderr/stdout for each benchmark run.

fitzgen:
CI won't pass here until we merge https://github.com/bytecodealliance/wasmtime/pull/2937

fitzgen:
> Subsequent PRs will add the checking of expected stderr/stdout for each benchmark run.

Actually this was really straightforward, so I've added it to this PR.

crates/recorder/src/bench_api.rs
```diff
@@ -0,0 +1,234 @@
...
+            wasm_bench_create: lib.get(b"wasm_bench_create")?,
+            wasm_bench_free: lib.get(b"wasm_bench_free")?,
+            wasm_bench_compile: lib.get(b"wasm_bench_compile")?,
+            wasm_bench_instantiate: lib.get(b"wasm_bench_instantiate")?,
+            wasm_bench_execute: lib.get(b"wasm_bench_execute")?,
+        })
+    }
+}
+
+/// An engine from a `BenchApin`.
```
abrown, line 67:
```suggestion
/// An engine from a `BenchApi`.
```

abrown (approved):

fitzgen:
Thanks!

PR #129, opened 2021-05-25
Add a `bz2` round tripping benchmark

fitzgen:
This benchmark is originally based on `bzip2` which has been adopted into
various places over the years: Valgrind's test suite, SPEC CPU, people's one-off
Wasm benchmarking, etc...

> This benchmark is basically bzip2 (mashed to be a single file) compressing and
> decompressing some data.  It tests Valgrind's handling of realistic and
> "difficult" (ie. lots of branches and memory accesses) integer code.
> Execution is spread out over quite a few basic blocks; --profile-flags
> indicates that to get to the top 90%th percentile of dynamic BB counts
> requires considering the top 51 basic blocks

The input was originally built up by the program, but I've modified it to read
input from the `./default.input` file. I've also made the default input a bit
larger by appending the source for a sizable Wikipedia page.


cfallin (approved):
LGTM!

Might be good to add a brief note at the top of `bz2.c` noting its most recent provenance ([link](https://github.com/cfallin/cranelift-new-backend-tests/blob/master/bz2/bz2.c)) -- I did make a few changes that weren't in the original (e.g. removing all uses of floating point, which was necessary at the time because we hadn't built that yet).

PR #130, opened 2021-05-26
Fix multi process measurements

### LINKED ISSUE #125: CSV headers are emitted N times when we use N processes
They should only be emitted once regardless how many processes we use when collecting samples.

fitzgen:
Previously we would emit N copies of the CSV headers for our N processes, and N
JSON blobs (rather than one combined JSON blob) for N processes. Now we parse
the output from the child process and accumulate all the results from all
processes, and finally dump this accumulation after all subprocesses are done.

This also adds smoke tests that check we can benchmark with multiple processes
and that the resulting CSV/JSON parses okay. These can serve as a starting point
for a richer test suite.

Fixes #125

### DIFF
--- a/.github/workflows/sightglass.yml
+++ b/.github/workflows/sightglass.yml
@@ -10,6 +10,10 @@ env:
   CARGO_TERM_COLOR: always
   RUST_LOG: info
 
+  # The Wasmtime commit that we build the bench API from for testing. Bumping
+  # this will automatically cause us to update our CI cache on the next run.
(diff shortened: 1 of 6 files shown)


abrown (approved):
Good call on the smoke test; looks like it exposed a few potential issues.

fitzgen:
Not sure why `sightglass-cli build-engine wasmtime` is failing in CI. Perhaps we need to install docker or something first.

PR #135, opened 2021-05-29
Use lazy-static to initialize Precision only once per process.

### LINKED ISSUE #134: Benchmarking initializes a `precision::Precision` for every run, spending a second or two in a calibration loop
I spent some time looking into why my benchmark runs seemed to be "choppy", sitting doing ~nothing between each (very fast) Wasm benchmark run. It seems that we have pauses of up to a few seconds between each run.

It turns out that this is because the `wall_cycle` measurement provider in the `recorder` crate invokes `precision::Precision::new`, building a new precision measurement object, for each run: [link](https://github.com/bytecodealliance/sightglass/blob/b465ad29d278a8eef3bc1fe2ba1db33892b7b57f/crates/recorder/src/measure/wall_cycles.rs#L12)

The `precision` crate notes in its [docs](https://docs.rs/precision/0.1.11/precision/struct.Precision.html) that "Note that on Linux system, this will perform calibration before returning" and recommends against running it more than necessary.

It seems we could get a significant speedup by saving one `Precision` object, maybe in a lazy-init'd global or somesuch, and reusing it.

cfallin:
Fixes #134.

### DIFF
--- a/crates/recorder/src/measure/wall_cycles.rs
+++ b/crates/recorder/src/measure/wall_cycles.rs
@@ -2,27 +2,33 @@
 //! wrapper around the `precision` crate to adapt it to the [Measure] API.
 
 use super::{Measure, Measurements};
+use lazy_static::lazy_static;
 use precision::{Config, Precision, Timestamp};
 use sightglass_data::Phase;
 
-pub struct WallCycleMeasure(Precision, Option<Timestamp>);
+lazy_static! {
+    static ref PRECISION: Precision = Precision::new(Config::default()).unwrap();
+}
+
+pub struct WallCycleMeasure(Option<Timestamp>);
 
 impl WallCycleMeasure {
     pub fn new() -> Self {
-        let precision = Precision::new(Config::default()).unwrap();
-        Self(precision, None)
+        Self(None)
     }
 }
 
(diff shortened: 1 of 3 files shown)


crates/recorder/src/measure/wall_cycles.rs
```diff
@@ -2,27 +2,34 @@
...
     fn start(&mut self) {
-        self.1 = Some(self.0.now());
+        let start = PRECISION.now();
+        self.0 = Some(start);
     }
 
     fn end(&mut self, phase: Phase, measurements: &mut Measurements) {
-        let end = self.0.now();
-        let elapsed = end - self.1.take().expect("an existing timestamp");
+        let precision = PRECISION.deref();
```
abrown, line 29: Why do we have to `deref()` here in `end` but not in `start`?

cfallin: Ah, so it's implicit in `start()` but I explicitly deref'd once here to avoid doing the lazy-init atomic access and branch twice. Though you have a good point in that the explicit trait call is a bit incongruent; I switched to the usual `&*PRECISION` idiom.

abrown (approved):

PR #137, opened 2021-06-01
Print a message when tests need to build a `libwasmtime_bench_api.so`

### LINKED ISSUE #136: CLI tests hanging
Not sure what happened, but the CLI tests are hanging for me on my local machine after I pulled the latest `main` branch. They used to take 10s to complete, which I thought was a little bit longer than they should, but I was hopeful that after #135 they would speed up, so I wanted to measure them again. Now they are just hanging. So I did a `git bisect` to find where they started hanging, and now they are hanging in the very first commit that introduced the CLI tests: 434512293fd1f5a358182122e8627dceaed469cf. This is perplexing, because they definitely didn't used to hang when I introduced that commit!

I've also verified that manually running the commands that the tests run does not hang locally.

I'm going to try profiling the tests to see where time is being spent now.

fitzgen:
This can take quite a while, and we don't want developers to mistakenly think that the tests are just hanging.

Fixes #136

### DIFF
--- a/crates/cli/tests/tests.rs
+++ b/crates/cli/tests/tests.rs
@@ -23,6 +23,24 @@ fn sightglass_cli_benchmark() -> Command {
         // system.
         static BUILD_WASMTIME: std::sync::Once = std::sync::Once::new();
         BUILD_WASMTIME.call_once(|| {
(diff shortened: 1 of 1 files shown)


cfallin (approved):
Good idea!

PR #144, opened 2021-06-04
Always use "faster" in effect size summary

### LINKED ISSUE #141: Faster/slower summary messages are confusing: invert one of the adjectives or one of the ratios
The `sightglass-cli benchmark` command produces output like the following:

```
  baseline.so is 1.00x to 1.03x FASTER than target/release/libwasmtime_bench_api.so!
  target/release/libwasmtime_bench_api.so is 0.97x to 1.00x SLOWER than baseline.so!
```

The intent to show the ratio in both directions is nice; however, these two statements actually contradict, I think. if A is n times faster than B, then B is n times slower than A. However, this summary says that B is 1/n times slower than A.

In other words, either both words above should be "FASTER", or the second ratio should also be 1.00x to 1.03x.

This was producing some confusion for me earlier and I had to read the source to figure out which direction is intended.

fitzgen:
A faster value less than 1 is actually slower, but to say "0.7x slower" isn't
correct, and actually means the opposite.

Fixes #141

### DIFF
--- a/crates/cli/src/benchmark.rs
+++ b/crates/cli/src/benchmark.rs
@@ -463,14 +463,9 @@ fn display_effect_size(
             let ratio_ci = effect_size.half_width_confidence_interval / effect_size.a_mean;
             writeln!(
                 output_file,
-                "  {a_engine} is {ratio_min:.2}x to {ratio_max:.2}x {faster_slower} than {b_engine}!",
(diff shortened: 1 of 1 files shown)


fitzgen:
The CI failure is fixed by the third commit in #143.

cfallin (approved):
Awesome, thanks!

PR #151, opened 2021-09-21
Add meshoptimizer

jlb6740:
Adds meshoptimizer's (https://github.com/zeux/meshoptimizer) codecbench benchmark.


benchmarks-next/meshoptimizer/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+WORKDIR /usr/src
+RUN git clone https://github.com/jlb6740/meshoptimizer.git \
+    && cd meshoptimizer \
+    && git checkout sightglass
+
+# Second, compile the benchmark to Wasm.
+
+WORKDIR /usr/src/meshoptimizer
+#COPY benchmark.c .
```
fitzgen, line 17: This can be removed, I think.

benchmarks-next/meshoptimizer/sightglass.h
```diff
@@ -0,0 +1,37 @@
+#ifndef sightglass_h
```
fitzgen, line 1: I think the other benchmarks symlink this file rather than copy it, that way we can update it for all benchmarks at once easily.

fitzgen (commented):
Can you add a README to the new benchmark's directory describing what it does?

Also, it would be great to move the input into a file and write the output to stdout so that the computation can't be optimized away by the compiler. And it would also be great to have an expected output file so that we can check that the program ran as expected, and therefore the benchmark results are valid.

See https://github.com/bytecodealliance/sightglass/tree/main/benchmarks-next#minimal-technical-requirements for details on the expected stdout stuff, and https://github.com/bytecodealliance/sightglass/blob/main/benchmarks-next/bz2/stdout.expected as an example.

jlb6740:
Hi @fitzgen I want to add meshoptimizer to a list of benchmarks in benchmark-next. I get this one CI error regarding a diff of wasm files not being the same. The job is as follows:

```
jobs:
  rebuild:
    runs-on: ubuntu-latest
    steps:e
    - uses: actions/checkout@v2
    - name: Install nightly
      uses: actions-rs/toolchain@v1
      with:
        toolchain: nightly
    - name: Rebuild benchmarks
      run: benchmarks-next/build.sh
    - name: Verify no benchmarks have changed
      run: git diff --no-ext-diff --name-only --exit-code
```
One benchmark appears to be a diff (not sure why) and the other is new. What is the means for modifying and adding new benchmarks in lieu of this check?

fitzgen:
It looks like these are the benchmarks that end up different after rebuilding:

* `benchmarks-next/meshoptimizer/benchmark.wasm`
* `benchmarks-next/pulldown-cmark/benchmark.wasm`

The first is interesting. Did you commit the most up-to-date build? Otherwise this points at nondeterministic builds, which isn't great. We should probably [attach the built files as artifacts](https://github.com/bytecodealliance/wasmtime/blob/main/.github/workflows/main.yml#L205-L209) from the CI job so that we can download them and commit them without having to actually rebuild locally ourselves. Then if the CI job fails on re-push we know that there is non-determinism rather than just out-of-date `.wasm` binaries committed.

I'm not sure why the `pulldown-cmark` benchmark is apparently being nondeterministic, but I think @abrown observed this before. I _think_ there might be a hash map used in the `build.rs` script for that crate or something. I think we can ignore it for now and investigate further in follow ups.

benchmarks-next/meshoptimizer/Dockerfile
jlb6740: > It looks like these are the benchmarks that end up different after rebuilding:
> 
> * `benchmarks-next/meshoptimizer/benchmark.wasm`
> * `benchmarks-next/pulldown-cmark/benchmark.wasm`
> 
> The first is interesting. Did you commit the most up-to-date build? Otherwise this points at nondeterministic builds, which isn't great. We should probably [attach the built files as artifacts](https://github.com/bytecodealliance/wasmtime/blob/main/.github/workflows/main.yml#L205-L209) from the CI job so that we can download them and commit them without having to actually rebuild locally ourselves. Then if the CI job fails on re-push we know that there is non-determinism rather than just out-of-date `.wasm` binaries committed.
> 
> I'm not sure why the `pulldown-cmark` benchmark is apparently being nondeterministic, but I think @abrown observed this before. I _think_ there might be a hash map used in the `build.rs` script for that crate or something. I think we can ignore it for now and investigate further in follow ups.

Ok, thanks! When I resubmit I'll see if these go away. Perhaps for meshoptimizer I did submit an older version. Really I just submitted this to see if it would pass CI so expected a few clean-up items.

jlb6740:
> Can you add a README to the new benchmark's directory describing what it does?
> 
> Also, it would be great to move the input into a file and write the output to stdout so that the computation can't be optimized away by the compiler. And it would also be great to have an expected output file so that we can check that the program ran as expected, and therefore the benchmark results are valid.
> 
> See https://github.com/bytecodealliance/sightglass/tree/main/benchmarks-next#minimal-technical-requirements for details on the expected stdout stuff, and https://github.com/bytecodealliance/sightglass/blob/main/benchmarks-next/bz2/stdout.expected as an example.

Hi @fitzgen sorry, been out since late last week. The code run is already a benchmark designed such that there shouldn't be concern for computation being stripped away. The input is already created at runtime: https://github.com/zeux/meshoptimizer/blob/master/tools/codecbench.cpp#L176-L213

For expected out I suppose we could compare certain lines from the verbose output:

> jlbirch@jlbirch[2021-09-28 13:24:14]:~/jlb6740-meshoptimizer$ codecbench -v | grep 'source\|data'
> source: vertex data 32064032 bytes, index data 24000000 bytes
> pass 0: vertex data 20521224 bytes, index data 2332680 bytes
> pass 1: vertex data 20522369 bytes, index data 2001016 bytes
> filters: oct8 data 32000000 bytes, oct12/quat12 data 64000000 bytes

There is no way then to give sightglass parameters or to run or tell it what from the output to grep ... these things need to be baked into the benchmark, correct?

fitzgen:
> The code run is already a benchmark designed such that there shouldn't be concern for computation being stripped away. The input is already created at runtime: https://github.com/zeux/meshoptimizer/blob/master/tools/codecbench.cpp#L176-L213

I don't see why this should stop an optimizing compiler from optimizing away the inputs. It can see that the benchmark function is called with exactly once, and then inline it into the same function that creates the benchmark input, and then do various optimizations to boil away the computation at compile time rather than run time.

I don't know whether (or to what degree) that is actually happening here, but in principle nothing is stopping it.

I don't think we necessarily need to address this ourselves. But it is something we should consider when deciding whether to include a particular benchmark. Is this a valuable enough along our other axis of benchmark desirability that this particular flaw is outweighed by other strengths? I'm assuming so? (It isn't clear to me what this benchmark does or why we think it is a good program to benchmark; this would be great info to include in a `README.md` in the benchmark's directory).

> There is no way then to give sightglass parameters or to run or tell it what from the output to grep ... these things need to be baked into the benchmark, correct?

The benchmark runner checks the program's full `stdout` output against the `stdout.expected` file.

Does this program dump too much stuff to `stdout`? That is also a warning sign, if it is printing things to `stdout` in the middle of the portion of the program that we are measuring between `bench.start` and `bench.end`, since we do *not* want to measure time spent doing I/O, only the compute itself.

jlb6740:
> > The code run is already a benchmark designed such that there shouldn't be concern for computation being stripped away. The input is already created at runtime: https://github.com/zeux/meshoptimizer/blob/master/tools/codecbench.cpp#L176-L213
> 
> I don't see why this should stop an optimizing compiler from optimizing away the inputs. It can see that the benchmark function is called with exactly once, and then inline it into the same function that creates the benchmark input, and then do various optimizations to boil away the computation at compile time rather than run time.
> 
> I don't know whether (or to what degree) that is actually happening here, but in principle nothing is stopping it.
> 
> I don't think we necessarily need to address this ourselves. But it is something we should consider when deciding whether to include a particular benchmark. Is this a valuable enough along our other axis of benchmark desirability that this particular flaw is outweighed by other strengths? I'm assuming so? (It isn't clear to me what this benchmark does or why we think it is a good program to benchmark; this would be great info to include in a `README.md` in the benchmark's directory).
> 

Yes, ok of course you are right that doing enough emulation through transformations can lead to pre-calculation and thus the removal of any code not dependent on compile time unknowns. I just didn't recognize that as a practical concern here .. we have a lot of other benchmarks here already that are in the same boat of not depending on input and to be consistent we'd need to change those as well. To be fair to this particular benchmark, it is coming from existing code where portability is probably a benefit and so not having to read in input I think probably could be viewed as an attribute more so than a flaw. But I'm sure we can probably modify this to read in the input and that maybe has the added benefit of being able to supply different input more easily if that matters.

> > There is no way then to give sightglass parameters or to run or tell it what from the output to grep ... these things need to be baked into the benchmark, correct?
> 
> The benchmark runner checks the program's full `stdout` output against the `stdout.expected` file.
> 
> Does this program dump too much stuff to `stdout`? That is also a warning sign, if it is printing things to `stdout` in the middle of the portion of the program that we are measuring between `bench.start` and `bench.end`, since we do _not_ want to measure time spent doing I/O, only the compute itself.

Out of the box this benchmark dumps the time/score only and does not dump the part I was trying to grep. We just need to modify this to print what we want and stop printing what we don't want. About the printing things to stdout in the middle of the program .. I believe the verbose is just printing out to see that input is as expected and is not the deterministic result of some calculation.

fitzgen:
> but there is still this issue related to peepmatic

What is the peepmatic issue?

fitzgen:
In the meantime, I think we can simply disable rebuilding the benchmarks in CI and checking that they are identical to the existing builds. It is rather unfortunate, but since the sources of non-determinism are largely outside of our control, and we don't want to *not* have these benchmarks, I think we can punt on the issue.

PR #153, opened 2022-01-13
recorder: upgrade precision crate

abrown:
This is an attempt to fix the build error observed in #152.


fitzgen (approved):

abrown:
I was hoping a version bump would fix this but it looks like I'll need to submit an upstream change.

abrown:
Thanks @jedisct1!

Issue #103, opened 2021-03-18
Support for setting different cranelift flags / wasmtime::Config flags


fitzgen:
It would be nice to benchmark these configurations (both one off and continuously over time):

* builds w/ vs w/out debug info
* optimized(speed) vs optimized(speed + size) vs not optimized
* bounds checking strategy (explicit vs virtual memory)
* with or without fuel

All of these things require dynamically setting flags on cranelift and/or wasmtime.

cfallin:
A proposal: we could extend the syntax for engine specifications (`wasmtime@branch@https://repo/path`) to include settings as well, taking inspiration from URLs' query strings. Perhaps something like:

`wasmtime@branch@https://github.com/my_fork_of/wasmtime?regalloc=linearscan&fuel=true&heapchecks=explicit`

We can bikeshed on the syntax -- no need to literally use the query-string format -- but it seems like having a single string that uniquely identifies (i) engine, (ii) which exact revision/source tree of that engine to use, and (iii) runtime configuration of that engine, would be very useful.

abrown:
@cfallin, I have a solution for the above on a branch in my private repository: [`BuildInfo`](https://github.com/abrown/sightglass/blob/fingerprint/crates/build/src/buildinfo.rs). It is a mechanism for both specifying the configuration of an engine when we built it and a way to capture that configuration to allow a reproducible build. I've found it helpful for uploading Sightglass data to the ES/Kibana server I'm using.

But before we can switch to that, we need a way to pass the runtime flags of the engine: I propose we do so as an opaque string and, on the Wasmtime bench API side, parse that string using the current CLI structopt parsing (this way there is no confusion when running benchmarks). So let me re-phrase what the problems are with this proposal, adapted from https://github.com/bytecodealliance/sightglass/pull/164#issuecomment-1097107532: in looking into this, we could use the structopt-enabled [`CommonOptions`](https://github.com/bytecodealliance/wasmtime/blob/main/src/lib.rs#L142) to transform a string into a [`Config`](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasmtime/src/config.rs#L83) by using the structopt `from_iter` function. The issue is that `CommonOptions` is private in the `wasmtime-cli` crate and `Config` lives in the `wasmtime` crate.

How could we use the same parsing as the `wasmtime-cli` crate, but in the `wasmtime-bench-api` crate? 
  a. make `CommonOptions` public so that `wasmtime-bench-api` can rely on it,
[...]

PR #164, opened 2022-04-12
Add a measurement mechanism for capturing phases in VTune

abrown:
This adds a `--measure vtune` option to the `benchmark` command that
informs the VTune engine of the various benchmarking phases
(compilation, instantiation, execution). This new measurement option
does not output any data (like the `noop` measure); instead, VTune
receives "task" markers that indicate when each phase starts and ends.
This facilitates performance analysis by allowing us to focus in on the
phase we're interested in.

An example of how to use this:
```text
vtune -collect hotspots sightglass-cli benchmark <path to>/benchmark.wasm --measure vtune
```

This change still has one major TODO:
 - the `wasmtime-bench-api` crate does not dump its JIT-compiled code,
   which is necessary for VTune to inspect the code in question; this
   functionality exists (see `ProfilingStrategy` in the [VTune docs])
   but we must find a way to configure this from Sightglass

[VTune docs]: https://docs.wasmtime.dev/examples-profiling-vtune.html#turn-on-vtune-support

### DIFF
--- a/crates/recorder/src/bench_api.rs
+++ b/crates/recorder/src/bench_api.rs
@@ -105,13 +105,13 @@ where
             stdin_path_ptr: stdin_path.as_ref().map_or(ptr::null(), |p| p.as_ptr()),
             stdin_path_len: stdin_path.as_ref().map_or(0, |p| p.len()),
             compilation_timer: measurement_data as *mut u8,
-            compilation_start: Self::start,
+            compilation_start: Self::compilation_start,
             compilation_end: Self::compilation_end,
             instantiation_timer: measurement_data as *mut u8,
-            instantiation_start: Self::start,
+            instantiation_start: Self::instantiation_start,
             instantiation_end: Self::instantiation_end,
             execution_timer: measurement_data as *mut u8,
-            execution_start: Self::start,
+            execution_start: Self::execution_start,
             execution_end: Self::execution_end,
         };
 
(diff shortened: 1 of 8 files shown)


abrown:
@fitzgen, re: the TODO: to be able to turn on the `--vtune` flag of Wasmtime, we need to be able to pass through some extra data in the bench API. If we could pass an `engine_flags` string and later parse it into a `Config` in Wasmtime, we could both fix this TODO and resolve #103, which wants a general way to configure the engine. In looking into this, we could use the structopt-enabled [`CommonOptions`](https://github.com/bytecodealliance/wasmtime/blob/main/src/lib.rs#L142) to transform a string into a [`Config`](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasmtime/src/config.rs#L83) by using the structopt `from_iter` function. The issue is that `CommonOptions` is private in the `wasmtime-cli` crate and `Config` lives in the `wasmtime` crate. There are solutions to this: a) make `CommonOptions` public so that `wasmtime-bench-api` can rely on it, b) make `Config` parse-able from a string (perhaps by moving `CommonOptions` somewhere closer but bringing in a structopt dependency). What are your thoughts on this? I like the "pass a engine flags string" approach because it doesn't require changes to Sightglass or the bench crate when configuration changes.

crates/recorder/src/bench_api.rs
```diff
@@ -138,11 +138,25 @@ where
...
         let measure = unsafe { data.as_mut().unwrap().0.as_mut().unwrap() };
-        measure.start();
+        measure.start(Phase::Compilation);
+    }
+
+    extern "C" fn instantiation_start(data: *mut u8) {
+        log::info!("Starting instantiation measurement");
+        let data = data as *mut (*mut M, *mut Measurements<'b>);
+        let measure = unsafe { data.as_mut().unwrap().0.as_mut().unwrap() };
+        measure.start(Phase::Compilation);
```
cfallin, line 152: Should this be `Instantiation` instead (copy/paste error)?

cfallin (approved):
LGTM, with one suspected typo below.

PR #166, opened 2022-04-21
Add a V8 engine

abrown:
This change adds the beginnings of a new V8 engine to Sightglass. It uses V8's `libwee8` library as the backing engine and constructs a `libengine.so` in C++ that is compatible with Sightglass. As-is, it has some limitations (discussed below), but the current change can run benchmarks under very specific circumstances and represents a significant effort to make this all work. I do not expect to include this as a part of CI until some of these limitations are resolved.

Limitations, in order of importance:
 - V8 expects only one engine to be instantiated in any process; to fix this, I would expect that we would need a safe way to create a static reference to the engine in the shared library (?). To get around this now, the example `make test-lib` will only run a single iteration
 - The WASI support for this library is minimal and merely stubbed out; this prevents allows running only the simplest of benchmarks. We should consider integrating an existing WASI implementation, e.g., uvwasi; the challenge is that some of the WASI calls are modified, i.e., to collect all stdout and stderr into files. To avoid this limitation for now, this change adds a `--no-output-check` to skip checking for collected output.
 - The `libwee8` build is slow: it takes approximately 27 minutes on my machine. Much of the time is spent retrieving the various dependencies of V8. It is not clear to me that all of these are needed and it may be possible to compile `libwee8` with less fetching.

For now, though, this library is functional:

```
$ make test-lib
RUST_LOG=trace cargo run -- benchmark ../../benchmarks-next/noop/benchmark.wasm \
        --engine build/libengine.so --processes 1 --iterations-per-process 1 \
        --skip-output-check
   Compiling sightglass-cli v0.1.0 (/home/abrown/Code/sightglass/crates/cli)
    Finished dev [unoptimized + debuginfo] target(s) in 1.77s
     Running `/home/abrown/Code/sightglass/target/debug/sightglass-cli benchmark ../../benchmarks-next/noop/benchmark.wasm --engine build/libengine.so --processes 1 --iterations-per-process 1 --skip-output-check`
 TRACE sightglass_cli > Executing command: Benchmark(BenchmarkCommand { engines: ["build/libengine.so"], processes: 1, iterations_per_process: 1, raw: false, output_format: Json, output_file: None, measure: WallCycles, small_workloads: false, working_dir: None, wasm_files: ["../../benchmarks-next/noop/benchmark.wasm"], stop_after_phase: None, significance_level: 0.01, skip_output_check: true })
 DEBUG sightglass_artifact::engine > Using already-built engine path: build/libengine.so
 INFO  sightglass_cli::benchmark   > Using benchmark engine: build/libengine.so
 INFO  sightglass_cli::benchmark   > Using Wasm benchmark: ../../benchmarks-next/noop/benchmark.wasm
 INFO  sightglass_cli::benchmark   > Using working directory: ../../benchmarks-next/noop
 DEBUG sightglass_cli::benchmark   > Wasm benchmark size: 3871 bytes
 INFO  sightglass_recorder::bench_api > Starting new measurement
 INFO  sightglass_recorder::bench_api > Finished measuring compilation
 INFO  sightglass_recorder::benchmark > Compiled successfully
 INFO  sightglass_recorder::bench_api > Starting new measurement
 INFO  sightglass_recorder::bench_api > Finished measuring instantiation
 INFO  sightglass_recorder::benchmark > Instantiated successfully
fd_write
bench_start
 INFO  sightglass_recorder::bench_api > Starting new measurement
 INFO  sightglass_recorder::bench_api > Finished measuring execution
bench_end
fd_write
proc_exit
 INFO  sightglass_recorder::benchmark > Executed successfully
compilation
  ../../benchmarks-next/noop/benchmark.wasm
    cycles
      [91751789 91751789.00 91751789] build/libengine.so
    nanoseconds
      [28786700 28786700.00 28786700] build/libengine.so
instantiation
  ../../benchmarks-next/noop/benchmark.wasm
    cycles
      [57176762 57176762.00 57176762] build/libengine.so
    nanoseconds
      [17938945 17938945.00 17938945] build/libengine.so
execution
  ../../benchmarks-next/noop/benchmark.wasm
    cycles
      [4219 4219.00 4219] build/libengine.so
    nanoseconds
      [1323 1323.00 1323] build/libengine.so
```


fitzgen:
> The `libwee8` build is slow: it takes approximately 27 minutes on my machine. Much of the time is spent retrieving the various dependencies of V8. It is not clear to me that all of these are needed and it may be possible to compile `libwee8` with less fetching.

Do they publish prebuilt `.so`s or something we could use, instead of building from source?

engines/v8/args.gn
```diff
@@ -0,0 +1,9 @@
+is_component_build = false
+use_custom_libcxx = false
+v8_enable_fast_mksnapshot = true
+v8_enable_i18n_support = false
+v8_use_external_startup_data = false
+is_debug = true
+symbol_level = 2
+v8_optimized_debug = false
```
fitzgen, line 8: Does this mean we are building V8 with debug symbols and/or without optimizations?

engines/v8/src/bench-api.cc
```diff
@@ -0,0 +1,103 @@
...
+#include "bench-state.hh"
+#include "link.hh"
+#include "wasm.hh"
+
+extern "C" {
+
+int wasm_bench_create(wasm_bench_config_t config, void** out_ptr) {
+  BenchState* st = new BenchState(Config::make(config));
+  st->engine = wasm::Engine::make();  // TODO cannot instantiate an EngineImpl
+                                      // multiple times.
```
fitzgen, line 19: Can you do something like

```c++
static wasm::Engine* engine;
if (engine == nullptr) {
    engine = wasm::Engine::make();
}
st->engine = engine;
```

here?

engines/v8/src/bench-state.hh
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * @brief The internal state maintained through a Sightglass benchmarking
+ * session.
+ */
+class BenchState {
+ public:
+  BenchState(Config config) : config(config) {}
+  Config config;
+  std::unique_ptr<wasm::Engine> engine;
```
fitzgen, line 66: And then this would have to become a raw pointer or reference instead of a `unique_ptr`.

engines/v8/src/link.cc
```diff
@@ -0,0 +1,116 @@
...
+  auto import_names = std::vector<ImportName>();
+  for (size_t i = 0; i < import_types.size(); ++i) {
+    wasm::ImportType* type = import_types[i].get();
+    auto module_str = std::string(type->module().get(), type->module().size());
+    auto name_str = std::string(type->name().get(), type->name().size());
+    // std::cerr << "Found import: " << module_str << " " << name_str <<
+    // std::endl;
+    import_names.push_back(ImportName(module_str, name_str));
+  }
+  // std::cerr << "Number of imports: " << import_names.size() << std::endl;
```
fitzgen, line 18: nitpick: delete this commented out logging or add a proper logging macro that can turn logging on/off at build time or via env var.

engines/v8/src/wasi.cc
```diff
@@ -0,0 +1,48 @@
+#include "wasi.hh"
+#include <cstdlib>
+#include <iostream>
+
+// WebAssembly type: (func (param i32))
+auto proc_exit(const wasm::Val args[], wasm::Val results[])
+    -> wasm::own<wasm::Trap> {
+  std::cerr << "proc_exit" << std::endl;
+  // exit(args[0].i32()); // TODO this cannot actually exit here; should trap?
```
fitzgen, line 9: Yes, WASI traps when you call `proc_exit`, and saves the exit code in some side state so the embedder can differentiate between an exit and another kind of trap.

engines/v8/src/wasi.cc
```diff
@@ -0,0 +1,48 @@
...
+  return nullptr;
+}
+
+// WebAssembly type: (func (param i32 i32 i32 i32) (result i32))
+auto fd_write(const wasm::Val args[], wasm::Val results[])
+    -> wasm::own<wasm::Trap> {
+  std::cerr << "fd_write" << std::endl;
+  // TODO actually write contents; needs access to linear memory.
+  // results[0] = args[0].copy();
+  results[0] = wasm::Val::i32(65553);
```
fitzgen, line 19: Why not return the length of the buffer, even if we didn't actually do the write (for now)

fitzgen (commented):
Can you add `engines/v8/README.md` that describes what this is, gives an overview of its current implementation status, and explains why it is here (rough comparison point for wasmtime/cranelift) and why it is not (benchmark wars, getting rigorous benchmark result numbers for V8)

PR #176, opened 2022-05-25
Add a `fingerprint` command

abrown:
This new command adds the ability to collect additional information
about several Sightglass components:
- `--kind machine` collects system information, e.g., OS, CPU, etc.
- `--kind benchmark` calculates a shortened benchmark name and hashes
  the Wasm contents
- `--kind engine` collects information about the Wasm engine, e.g., how
  it was built and what flags it uses

### DIFF
--- a/crates/fingerprint/src/benchmark.rs
+++ b/crates/fingerprint/src/benchmark.rs
@@ -1,5 +1,5 @@
 use crate::hash;
-use crate::util::stringify;
+use crate::util::to_string_lossy;
 use anyhow::{anyhow, Result};
 use serde::{Deserialize, Serialize};
 use std::path::PathBuf;
@@ -65,7 +65,7 @@ fn simplify_benchmark_name<P: AsRef<Path>>(path: P) -> Result<String> {
     } else {
         stem
     };
-    Ok(stringify(name))
+    Ok(to_string_lossy(name))
 }
 
 /// Simplify the benchmark path if possible; e.g.:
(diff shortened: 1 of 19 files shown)


crates/fingerprint/src/hash.rs
```diff
@@ -0,0 +1,35 @@
...
+pub(crate) fn hexify(bytes: &[u8]) -> String {
+    use std::fmt::Write;
+    let mut s = String::new();
+    for byte in bytes {
+        write!(&mut s, "{:x}", byte).expect("unable to write byte as hex");
+    }
+    s
+}
+
+/// Create a hexadecimal string from a sequence of bytes.
```
fitzgen, line 32: This doc comment is copy pasted from above.

crates/fingerprint/src/util.rs
```diff
@@ -0,0 +1,6 @@
+use std::ffi::OsStr;
+
+/// Provide a common way to create `String`s from `OsStr`.
+pub(crate) fn stringify<S: AsRef<OsStr>>(s: S) -> String {
```
fitzgen, line 4: Can we name this `to_string_lossy` or something to capture the lossiness?

fitzgen (approved):
LGTM w/ a couple nitpicks below.

crates/fingerprint/src/benchmark.rs
```diff
@@ -0,0 +1,117 @@
...
+        );
+        assert_eq!(
+            simplify_benchmark_path("code/benchmarks-next/noop.wasm"),
+            format!("benchmarks-next{}noop.wasm", std::path::MAIN_SEPARATOR)
+        );
+        // Note here how `simplify_benchmark_path` does not modify the path separator.
+        assert_eq!(
+            simplify_benchmark_path("some/other/path.wasm"),
+            "some/other/path.wasm"
+        );
```
abrown, line 115: For the future: this is rather unexpected behavior, in that it isn't always clear when fingerprinting will use the OS path separator and when it will just pass it through. I'm going with "use the OS separator" instead of "replace everything to forward slashes" but I can see how in the future we might want to just use forward slashes to make comparing these things easier. Leaving this for later...

PR #177, opened 2022-05-25
Replace `build-benchmark` command with a Bash script

abrown:
With #170, the only user of the Docker wrapper in the `sightglass-build`
crate was the `build-benchmark` command provided by the
`sightglass-cli`. In order to reduce the amount of code to maintain, the
Docker wrapper and its CLI command, `build-benchmark`, are removed. They
are replaced by several Bash scripts:
 - `build.sh` now does the same work that `build-benchmark` did
   (building a Docker image, extracting a file, etc.)
 - `build-all.sh` is a renaming of the old `build.sh` script; it builds
   all the benchmarks in the directory (or a random subset if a numeric
   parameter is provided)
 - `clean.sh` is a new script to remove any temporary files, images, and
   containers
 - `run-all.sh` is a renaming of `run.sh`


.github/workflows/benchmarks.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       with:
         toolchain: nightly
     - name: Rebuild benchmarks
-      run: benchmarks-next/build.sh
+      run: benchmarks-next/build-all.sh 5
```
abrown, line 23: At this point, I don't think we need to build all of the benchmarks all the time. The purpose of this is to check that any benchmarks in the directory can indeed be built and by picking 5 of them in each CI run, its probable that we will find any newly-added benchmarks that don't build for some reason.

abrown: (context: rebuilding the benchmarks is taking ~35 minutes of CI time; the next closest job is running them, which takes ~10 minutes).

fitzgen (approved):
Thanks!

PR #182, opened 2022-06-16
Port the ability to pin to a single core from the legacy runner

abrown:
In the legacy runner, @jedisct1 added the ability to pin execution to a
single core, using several different mechanisms: it uses the `core_affinity`
crate for common OSes and the `hwloc` crate otherwise. This change ports
that functionality over to the new runner under the `--pin` flag.

This is the last feature that existed in the legacy runner but not in
the new runner (that I could see, at least). Though core-pinning can be
accomplished otherwise, e.g, `taskset --cpu-list 0`, these mechanisms
apply to all subprocesses. It could be possible that a user wants to
benchmark using multiple processes, all on potentially different CPUs,
but have all iterations in that process be pinned to a single core to
avoid measuring context switches.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1635,6 +1635,8 @@ name = "sightglass-recorder"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "core_affinity",
+ "hwloc",
  "ittapi",
  "lazy_static",
  "libc",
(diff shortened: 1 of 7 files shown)


jedisct1:
> Though core-pinning can be accomplished otherwise, e.g, `taskset --cpu-list 0`

This is not the case on Windows and macOS, which is the main reason this option was introduced.

So, yes, having this on the new running is not a bad idea :)

Issue #178, opened 2022-05-26
Weird file CI issues with benchmark_effect_size test


abrown:
On Windows, the `benchmark_effect_size` test has started intermittently failing since #170. It is unclear to me what could have modified the execution of the code that copies the built engine to a location provided by `tempfile::NamedTempFile`. I added an assert in hopes that it would trigger the error sooner but the OS always thinks the file exists:

https://github.com/bytecodealliance/sightglass/blob/cd8c11b1da4c7ebe3b5b1c7fc10e3a613a9c3b7f/crates/cli/tests/tests.rs#L213

One example of this failure is [here](https://github.com/bytecodealliance/sightglass/runs/6613565251?check_suite_focus=true#step:7:357):

```
thread 'benchmark::benchmark_effect_size' panicked at 'Unexpected failure.
code--1073741819
...
command=`"D:\\a\\sightglass\\sightglass\\target\\debug\\sightglass-cli.exe" "benchmark" "--engine" "\\\\?\\D:\\a\\sightglass\\sightglass\\engines\\wasmtime\\engine.dll" "--engine" "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\.tmpaNVVum" "--processes" "1" "--iterations-per-process" "3" "../../benchmarks-next/noop/benchmark.wasm"`
```

If [this random link](https://docs.microsoft.com/en-us/answers/questions/379441/error-code-34exited-with-code-107374181934.html) is to be believed, the code referenced above is an access violation that might indicate referencing a null pointer, e.g. But why? The original engine library seems to run fine according to the logs; only the copied engine library has the problem.

abrown:
@fitzgen, should we now consider this closed? If the CI issues with `benchmark_effect_size` are no longer a problem, maybe my "access violation" hypothesis was incorrect and it was the variance issue you saw?

fitzgen:
Yeah sure, we can always reopen if we see it again.

PR #183, opened 2022-06-17
Remove legacy runner

abrown:
This PR is a series of commits that remove the legacy runner from the Sightglass repository. With #182 now merged, the next version of the runner covers all known features of the legacy runner and adds a few that we have found useful since the switch over. Since the legacy runner is no longer actively maintained, this change removes that code and all of the associated legacy benchmarks. See each commit message for a record of this re-factoring.


fitzgen (approved):

abrown:
For the record: I've taken five stabs at re-running the failed `benchmark_effect_size` test. It continues to fail with the same problem described in #178. I believe that failure is completely unrelated to these changes.

PR #185, opened 2022-06-22
Fix benchmark commands in README

yurydelendik:
As is, the commands will do nothing. Switching "--engines" with benchmark names will fix that.


abrown:
The CI failure is unrelated to this change; I think it is another instance of the intermittent #178.

PR #188, opened 2022-06-24
Update wasmparser to 0.86

yurydelendik:
I was not able to compile some of the new benchmarks I was trying to submit. I think it was due to new (SIMD?) instructions. Updating of wasmparser helped.

P.S. weird modifications in cargo.lock

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,48 +3,12 @@
 version = 3
 
 [[package]]
-name = "addr2line"
-version = "0.17.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


abrown:
Maybe let's merge #187 first... Might resolve some of that weirdness.

abrown:
Ok, #187 is merged... I'm surprised this doesn't need a rebase?

yurydelendik:
> Ok, #187 is merged... I'm surprised this doesn't need a rebase?

Perhaps it is doing the same changed to the lock file.

abrown:
I think you're right, from glancing through the diff.

abrown (approved):

PR #193, opened 2022-08-05
Disable wall time measurement in "wall-cycles" measure; only measure cycles

fitzgen:
Rather than the `precision` crate, which is way too over-eager with its
calibration, and often takes 5 seconds to calibrate. This in turn was making
multi-process benchmark measurements way too slow to gather.

`minstant` does not suffer from these problems.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -772,9 +772,9 @@ checksum = "eb9f9e6e233e5c4a35559a617bf40a4ec447db2e84c20b55a6f83167b7e57872"
 
 [[package]]
 name = "precision"
-version = "0.1.14"
+version = "0.1.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "00d13bf03fb02ea18f91430a30f2e53dd7f064445400f16c0d97b17aa2791c2d"
+checksum = "bbc84fc8369453e02d82fe292dc9b03151c80a36716718c0dcc1ff0fe4029a63"
 dependencies = [
  "cc",
  "libc",
--- a/crates/recorder/Cargo.toml
+++ b/crates/recorder/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2021"
 anyhow = "1.0"
 libloading = "0.7"
 log = "0.4"
-precision = "0.1"
+precision = "0.1.15"
 serde = { version = "1.0.118", features = ["derive"] }
 sightglass-build = { path = "../build" }
 sightglass-data = { path = "../data" }
--- a/crates/recorder/src/measure/wall_cycles.rs
+++ b/crates/recorder/src/measure/wall_cycles.rs
@@ -7,7 +7,7 @@ use precision::{Config, Precision, Timestamp};
 use sightglass_data::Phase;
 
 lazy_static! {
-    static ref PRECISION: Precision = Precision::new(Config::default()).unwrap();
+    static ref PRECISION: Precision = Precision::new(Config::default().wall_time(false)).unwrap();
 }
 
 pub struct WallCycleMeasure(Option<Timestamp>);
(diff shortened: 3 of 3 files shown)


abrown (approved):
I'm a bit worried about the possibility of dropping down to coarse time unwittingly (having looked through the new crate) but this should fix the calibration annoyances. This crate also seems to be a bit Linux-specific... which we would measure everywhere with similar precision (I wonder what criterion ends up using?).

fitzgen:
At the end of [this line](https://github.com/bytecodealliance/sightglass/runs/7735912067?check_suite_focus=true#step:7:253):

> Error: The variance of one of the samples is zero

Can't remember if I rebased this on top of the PR to bump iterations up to 100. Will double check.

fitzgen:
If this doesn't fix it, we can either:

* try a different benchmark, instead of noop, that still runs relatively fast, in the hopes of getting more variance, or
* we can fix/fork upstream to return a proper error when there is no variance rather than `unwrap`

abrown:
> The calibration time in `precision` can be configured (this is the `setup_duration` value of the `Config` struct).
> 
> And I'd be glad to make any changes to that crate.

Ah, good to know. I've wondered in the past if, since converting cycles to nanoseconds is a bit of a fraught topic, perhaps we could add a feature to `precision` to only report cycles?

jedisct1:
> > The calibration time in `precision` can be configured (this is the `setup_duration` value of the `Config` struct).
> > And I'd be glad to make any changes to that crate.
> 
> Ah, good to know. I've wondered in the past if, since converting cycles to nanoseconds is a bit of a fraught topic, perhaps we could add a feature to `precision` to only report cycles?

Cycles cannot be counted on every platform (WebAssembly without JavaScript being a major example). But sure, that can totally be added.

fitzgen:
I tried lowering the warm up time to 500ms, which I think is still way too slow for a per-process overhead for our purposes, and now I get divide by zeros inside `Precision::guess_frequency_with_wall_clock`. 

But IIUC, the frequency stuff is just there to get nanoseconds, right? We would prefer to just have cycle counts, and we don't care about Wasm targets because Wasmtime doesn't itself run on Wasm.

I'll look into a PR for just getting cycles out of `precision`.

fitzgen:
Okay, that PR was merged, a new release of the `precision` crate published, and I've updated this PR accordingly! Want to take another look @abrown?

abrown (approved):
Yeah, I think this makes more sense. The `ns` measurement was always a bit suspect to me. Hopefully the benchmarks run significantly faster for you?

abrown (commented):
One other comment: perhaps the name of this measurement, `wall-cycles`, should just get shortened down to `cycles` in all of the CLI-facing code? Not necessary in this PR, though.

fitzgen:
Yep, a lot faster for multi-process measurements now! 100 processes x 1 iteration-per-process is now in the same ballpark as 1 process x 100 iterations-per-process:

```
$ time cargo run --release -- benchmark --stop-after compilation -e ~/wasmtime/target/release/libwasmtime_bench_api.so --processes 1 --iterations-per-process 100 -- benchmarks/noop/benchmark.wasm
    Finished release [optimized] target(s) in 0.05s
     Running `target/release/sightglass-cli benchmark --stop-after compilation -e /home/nick/wasmtime/target/release/libwasmtime_bench_api.so --processes 1 --iterations-per-process 100 -- benchmarks/noop/benchmark.wasm`
compilation
  benchmarks/noop/benchmark.wasm
    cycles
      [4450612 5430009.62 9096456] /home/nick/wasmtime/target/release/libwasmtime_bench_api.so

real	0m0.301s
user	0m1.038s
sys	0m0.161s

$ time cargo run --release -- benchmark --stop-after compilation -e ~/wasmtime/target/release/libwasmtime_bench_api.so --processes 100 --iterations-per-process 1 -- benchmarks/noop/benchmark.wasm
    Finished release [optimized] target(s) in 0.05s
     Running `target/release/sightglass-cli benchmark --stop-after compilation -e /home/nick/wasmtime/target/release/libwasmtime_bench_api.so --processes 100 --iterations-per-process 1 -- benchmarks/noop/benchmark.wasm`
compilation
  benchmarks/noop/benchmark.wasm
    cycles
      [6251221 7131495.31 10990615] /home/nick/wasmtime/target/release/libwasmtime_bench_api.so

real	0m0.547s
user	0m1.076s
sys	0m0.607s
```

PR #195, opened 2022-08-08
Make `benchmark_effect_size` test less flaky

fitzgen:
It complains if there isn't any variance between samples, so bump up the number
of iterations we perform to make getting all identical samples extremely
unlikely.

This flaky test seems to be holding back #194 and #193

### DIFF
--- a/crates/cli/tests/all/benchmark.rs
+++ b/crates/cli/tests/all/benchmark.rs
@@ -146,7 +146,7 @@ fn benchmark_effect_size() -> anyhow::Result<()> {
         .arg("--processes")
         .arg("1")
         .arg("--iterations-per-process")
-        .arg("3")
+        .arg("200")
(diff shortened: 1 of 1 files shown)


abrown:
Wait... how did you figure out that variance is the issue with this test? When I look at the test output, I don't see that, e.g., with this [latest failure](https://github.com/bytecodealliance/sightglass/runs/7730718273?check_suite_focus=true#step:7:255). I guessed that it was some weird file issue in #178 but if it's actually fixed by more runs, then great! But how? Shouldn't we be able to see something of this in the test output?

fitzgen:
I think the crate we use has some internal unwraps/panics rather than exposing results.

But I did see at least one bit of log output for one of the failures (maybe before doing a retry?) that mentioned variance.

abrown (approved):
Well, let's see if this works!

PR #201, opened 2022-09-15
Adding a benchmark for wasi-nn

brianjjones:
This benchmark will measure the performance of
wasi-nn by measuring the time required to complete a single inference.


brianjjones:
One issue that needs to be addressed is that this benchmark requires OpenVINO to run. So I would think it would be best for it to be an optional benchmark that doesn't get executed automatically by run-all.sh. Thoughts on the best course of action?

brianjjones:
@abrown This PR is ready for review now. It's been updated with the changes we discussed. The one possible issue is that the current build.sh script expects the file created in Docker to be named `benchmark.wasm`. Perhaps we could change this to look for any file that matches `*-benchmark.wasm`? This way, when build_all.sh or build.sh is called on image-classification, it names it image-classification-benchmark.wasm instead. This prevents run_all.sh from accidentally picking up the image-classification benchmark.

benchmarks/image-classification/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM rust:1.54
+
+RUN rustup target add wasm32-wasi
+WORKDIR /usr/src
+ADD rust-benchmark rust-benchmark
+WORKDIR /usr/src/rust-benchmark
+RUN cargo build --release --target=wasm32-wasi
+RUN cp target/wasm32-wasi/release/image-classification-benchmark.wasm /benchmark.wasm
```
abrown, line 8: Yeah, I agree with your comment above about changing the `build.sh` to be a more liberal "copier." The problem is that I'm not sure you can use glob patterns like `*benchmark.wasm` with Docker commands? If you can, then great, we should do that; otherwise we may have to modify `build.sh` to copy everything from some directory, e.g., `/sightglass-output`.

benchmarks/image-classification/setup.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# This will download a copy of OpenVINO and extract it locally. Once this script is run, you can use this command to run it:
+# cargo run -- benchmark benchmarks/wasi-nn/benchmark.wasm --engine engines/wasmtime/libengine.so
+
+WASI_NN_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+FILENAME=l_openvino_toolkit_ubuntu20_2022.2.0.7713.af16ea1d79a_x86_64
+[ ! -d ${WASI_NN_DIR}/openvino ] && wget -nc -q --no-check-certificate https://storage.openvinotoolkit.org/repositories/openvino/packages/2022.2/linux/${FILENAME}.tgz -O ${WASI_NN_DIR}/${FILENAME}.tgz
+[ ! -d ${WASI_NN_DIR}/openvino ] && tar -C ${WASI_NN_DIR} -zxf ${WASI_NN_DIR}/${FILENAME}.tgz  && mv ${WASI_NN_DIR}/${FILENAME} ${WASI_NN_DIR}/openvino || echo "OpenVINO is already there, skipping..."
+source ${WASI_NN_DIR}/openvino/setupvars.sh
```
abrown, line 10: Could this script also retrieve `mobilenet.bin`, `mobilenet.xml`, and the `test.jpg`? The additional weight isn't a lot for those files (i.e., <14MB) but it might be nice to avoid downloading it if users don't want to.

engines/wasmtime/build.rs
```diff
@@ -66,7 +66,7 @@ fn main() {
     // Build the engine library.
     section("Building the engine");
     exec(
-        &["cargo", "build", "--release", "-p", "wasmtime-bench-api"],
+        &["cargo", "build", "--release", "-p", "wasmtime-bench-api", "--features", "wasi-nn"],
```
abrown, line 69: Did we figure out if this was actually necessary? I think it would be nicer (but this is a nit, not a real complained) if it could be removed either a) because it is unnecessary or b) by an upstream change to build in wasi-nn by default to `bench-api`.

benchmarks/image-classification/rust-benchmark/src/main.rs
```diff
@@ -0,0 +1,103 @@
...
+        .collect();
+    results.sort_by(|a, b| b.1.partial_cmp(&a.1).unwrap());
+    results
+}
+
+// Take the image located at 'path', open it, resize it to height x width, and then converts
+// the pixel precision to FP32. The resulting BGR pixel vector is then returned.
+fn image_to_tensor(path: String, height: u32, width: u32) -> Vec<u8> {
+    let pixels = Reader::open(path).unwrap().decode().unwrap();
+    let dyn_img: DynamicImage = pixels.resize_exact(width, height, image::imageops::Triangle);
```
abrown, line 82: Is there an issue anywhere (or TODO) to remind us to replace with the new and improved `image2tensor` crate?

benchmarks/image-classification/rust-benchmark/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "image-classification-benchmark"
+version = "0.1.0"
+authors = ["The Bytecode Alliance Developers"]
+readme = "README.md"
+edition = "2018"
+publish = false
+
+[dependencies]
+image = { version = "0.23.14", default-features = false, features = ["gif", "jpeg", "ico", "png", "pnm", "tga", "tiff", "webp", "bmp", "hdr", "dxt", "dds", "farbfeld"]  }
```
abrown, line 10: I think we're only using `"jpeg"` here, right?

benchmarks/image-classification/Dockerfile
abrown: I'm actually surprised by this question: how does the "rebuild" CI not fail? It seems like it should because this `Dockerfile` emits a `benchmark.wasm` but you have checked in a `image-classification-benchmark.wasm` &mdash; that should result in a failure when CI realizes we are not building the same files.

I think the answer is that `build-all.sh` picks a random subset of benchmarks to build, so this will eventually fail, just not every time. (Rebuilding the benchmarks takes a while so in an effort to speed up CI we just do a subset). You might want to run `build.sh` on this benchmark and see what happens.

(On a side note, I think at some point there was a Git check to make sure none of the bytes had changed, e.g., `git diff --exit-code`. That is no longer there so let me open an issue...)

abrown (commented):
I think this mostly looks good minus the comments I made below. I am interested also to hear @fitzgen's opinion on this PR as well.

benchmarks/image-classification/setup.sh
brianjjones: Sure, I guess I was just doing it this way because all the other benchmarks included everything needed to just run the provided benchmark.wasm. But since this requires you run setup.sh regardless, I don't see why they can't be downloaded.

engines/wasmtime/build.rs
brianjjones: Yes this isn't needed, I'll take it out.

benchmarks/image-classification/rust-benchmark/Cargo.toml
brianjjones: I suppose there's no reason currently to pull in the other image types.

benchmarks/image-classification/rust-benchmark/src/main.rs
brianjjones: I've added a TODO:

benchmarks/image-classification/setup.sh
brianjjones: Update: I'm keeping the original test.jpg because the old .bgr file doesn't seem to work with the example.

benchmarks/image-classification/Dockerfile
brianjjones: @abrown Not sure about why the CI didn't fail, but I just made some changes so now build.sh allows for different wasm file names besides benchmark.wasm. Now you can either pass the filename.wasm to build.sh, or it will look in the benchmark you are building for the name and use that.

jlb6740:
Hi @brianjjones 
Without being sure of the steps to follow to run this I did this:
1) cd benchmarks; ./build.sh image-classification/
2) cc image-classification ; "Read the readme" ; ./setup.sh
3) cc ../../ ; cargo run -- benchmark --engine engines/wasmtime/libengine.so --engine-flags=--wasi-modules=experimental-wasi-nn -- benchmarks/image-classification/ # but no wasm file.

Realized I should run ./build.sh after I ran.
1) cd benchmarks; ./build.sh image-classification/
2) cc../ ; cargo run -- benchmark --engine engines/wasmtime/libengine.so --engine-flags=--wasi-modules=experimental-wasi-nn -- benchmarks/image-classification/image-classification-benchmark.wasm 

But here there is this error:

>      Running `target/debug/sightglass-cli benchmark --engine engines/wasmtime/libengine.so --engine-flags=--wasi-modules=experimental-wasi-nn -- benchmarks/image-classification/image-classification-benchmark.wasm`
> Host error: BackendError(BackendAccess(library loading error
> 
> Caused by:
>     system failed to load shared libraries (see https://github.com/intel/openvino-rs/blob/main/crates/openvino-finder): the shared library at /opt/intel/openvino_2021/deployment_tools/inference_engine/lib/intel64/libinference_engine_c_api.so could not be opened: libinference_engine.so: cannot open shared object file: No such file or directory))

What are the correct steps?

brianjjones:
@jlb6740 For whatever reason sourcing the setupvars.sh script from setup.sh doesn't work. So you'll need to call the following:

```
source benchmarks/image-classification/openvino/setupvars.sh
```
After that, your cargo run should work.

brianjjones:
Also I'll update the README with more clear instructions.

brianjjones:
@jlb6740 I updated the instructions in the README, let me know if that's more clear and works for you.

jlb6740:
Looks good to me. Thanks for the updates.

jlb6740 (approved):

PR #204, opened 2022-10-10
Organize benchmarks into suites

### LINKED ISSUE #202: Organize benchmarks into suites
I am proposing in this issue that we add a way for `sightglass-cli` to benchmark a specific set of files together.

### Why

- I have noticed that certain users only run the larger benchmarks (`spidermonkey`, `bz2`, `pulldown-cmark`); in the past I have wanted to only run the `shootout-*` micro-benchmarks or "all the benchmarks that use SIMD."
- New benchmarks, like the wasi-nn one submitted in #201, would require special setup and would not run successfully with `sightglass-cli benchmark ... benchmarks/*/benchmark.wasm` (as the README currently advises the user). These special benchmarks (e.g., due to environment setup, compile-time or runtime flags, etc.) could be separated in to their own suite.
- As described in the RFCs on benchmarking, [candidate benchmarks](https://github.com/bytecodealliance/rfcs/blob/main/accepted/benchmark-suite.md#candidate-program-requirements) should be "real, widely used programs, or at least extracted kernels of such programs." Though this is good criteria, it can be subjective (e.g., though some shootout benchmarks are "extracted kernels" some may consider them too small to be representative); organizing the benchmarks into suites would recognize and document these differences.
- #71 requests a default set of benchmarks to run automatically if no benchmark files are specified &mdash; one could imagine a _default_ suite for that.

### How

The core idea would be to create a set of `*.suite` files. `*.suite` files would contain a newline-delimited list of benchmark paths relative to the Sightglass project directory and would accept `#` line comments. E.g., `foo.suite` could look like:

```
# This suite contains benchmarks that call the `foo` function.
benchmarks/foo/benchmark.wasm
benchmarks/bar/benchmark.wasm
benchmarks/baz/specially-crafted-benchmark.wasm
```

`sightglass-cli` would accept a new flag &mdash; `--suite` &mdash; with the path to the `*.suite` file to use. It would be an error to specify both a suite and a benchmark file; other than that, running benchmarks by path (the current way) would remain unchanged. Users could create their own `*.suite` files and manage these outside the repository but the Sightglass repository would contain a few `*.suite` files. There are many ways to slice this, but I could see the following initial set:

- `default.suite`
- `shootout.suite`
- `simd.suite`

If no benchmark paths or suites are specified, `sightglass-cli` would run the benchmarks contained in the `default.suite`.

-----

I would appreciate feedback on this proposal before implementing it. Any thoughts one way or another are appreciated!

abrown:
This change closes #202 by adding a way to group benchmarks into `*.suite` files. This incorporates @jameysharp's comments, e.g., to calculate the benchmark path relative to the suite path and to detect suite files based on their `.suite` extension.

Though more suites could be added in the future, the result of this is that suite files can be built (see `benchmarks/shootout.suite`, e.g.) and then run like:

```console
$ sightglass-cli benchmark benchmarks/shootout.suite -e ...
```


crates/cli/src/benchmark.rs
```diff
@@ -308,7 +314,7 @@ impl BenchmarkCommand {
             // and therefore potentially invalidating relative paths used here).
             let engine = check_engine_path(engine)?;
 
-            for wasm in &self.wasm_files {
+            for wasm in &self.benchmarks {
```
abrown, line 317: Not sure this looks right: I feel like we should be pushing the actual benchmark paths into `choices`, not entire suites.

crates/cli/src/suite.rs
```diff
@@ -0,0 +1,139 @@
...
+        Self::parse_contents(suite_path.as_ref(), &fs::read(suite_path.as_ref())?)
+    }
+
+    /// Utility function for easier testing.
+    fn parse_contents<P: AsRef<Path>>(suite_path: P, file_contents: &[u8]) -> Result<Self> {
+        use io::BufRead;
+        let suite_path = suite_path.as_ref().to_path_buf();
+        let parent_dir = suite_path
+            .parent()
+            .expect("the suite path must have a parent directory");
```
fitzgen, line 82: I think this should return an `Err` rather than panic.

crates/cli/src/suite.rs
```diff
@@ -0,0 +1,139 @@
...
+        let mut benchmarks = vec![];
+        for line in io::BufReader::new(file_contents)
+            .lines()
+            .filter_map(Result::ok)
```
fitzgen, line 86: This will silently ignore any errors. Instead we should do the (arguably uglier) thing where we `?`-propagate the error at the start of the loop body:

```rust
for line in io::BufReader::new(file_contents).lines() {
    let line = line?;
    // ...
}
```

crates/cli/src/suite.rs
```diff
@@ -0,0 +1,139 @@
...
+
+        assert_eq!(
+            suite.benchmarks,
+            vec![
+                // Relative paths are appended.
+                PathBuf::from("/home/a.wasm"),
+                // Absolute paths are preserved.
+                PathBuf::from("/b.wasm"),
+                // Canonicalization happens later.
+                PathBuf::from("/home/../c.wasm")
```
fitzgen, line 135: FWIW, we shouldn't ever call `canonicalize` on the path because it just does super weird stuff on windows that isn't really what we want, and using non-canonical paths doesn't hurt for our use cases in sightglass.

fitzgen (approved):
LGTM with comments addressed. Thanks for pushing this through!

crates/cli/src/suite.rs
abrown: If a suite is a file, it should be contained in a directory, right? I thought this error case was too uncommon to force the caller to figure out what to do with it?

crates/cli/src/benchmark.rs
abrown: What I ended up doing here is using `flat_map` so that each subprocess in `execute_in_multiple_processes` gets a single benchmark to run instead of potentially receiving a whole suite. `execute_in_current_process` stays the same in that it can execute either a suite or a single benchmark.

> We also have the `wasm_files` available in scope already (although they are `String`s and should probably be `PathBuf`s instead) that we can reuse for this loop.

We don't have `wasm_files` in this function `execute_in_multiple_processes`, only in `execute_in_current_process`. I did try, however, to see if I could use `PathBuf`s instead of `String`s but the fly in the ointment is `Measurements`: it wants a `&'a str` to refer to the Wasm file and I would guess `wasm_files` became a `Vec<String>` to satisfy that lifetime. There is probably a refactoring here to get to, e.g., `&Path` but I'll leave it as-is for this PR.

crates/cli/src/suite.rs
abrown: I made a note of this in the documentation just to remember why to keep these paths non-canonicalized. I chose to keep them non-canonicalized out of testing laziness; this is a more valid concern.

crates/cli/src/suite.rs
fitzgen: Ah okay I misunderstood.

Maybe clarify in the message something like

> suites are always files and files always have parents

and also `debug_assert!(suite_path.is_file())` or something since this function expects that and that invariant is maintained by callers rather than this function itself.

abrown: (Had to `debug_assert!` in `parse` instead of `parse_contents` to avoid creating files during testing... seems like an OK trade-off).

PR #205, opened 2022-10-12
Document low-variance benchmarking config

jameysharp:
I promised @abrown I'd write this procedure up, so here's a quick draft.


fitzgen (approved):
LGTM, thanks for the write up!

PR #208, opened 2022-10-20
Change the `spidermonkey.wasm` benchmark's JS workload to be a markdown parser

fitzgen:
This is a more realistic JS workload than computing the Nth fibonacci.

r? @cfallin


cfallin (approved):
LGTM! Nick and I pair-developed this yesterday; happy to see a more realistic JS workload in our suite. Thanks!

fitzgen:
Force push is just me adding

```
Co-Authored-By: Chris Fallin <chris@cfallin.org>
```

to the git commit :)

benchmarks/spidermonkey/sightglass.h
```diff
@@ -1 +0,0 @@
-../../include/sightglass.h
```
abrown, line 1: We probably want to keep the symlink in case we change this header in the future? Or are there changes to this file?

abrown (approved):
LGTM with the symlink comment.

benchmarks/spidermonkey/sightglass.h
fitzgen: The docker build was failing with the symlink. No idea why!

Issue #203, opened 2022-10-06
Emit measurement results immediately


abrown:
@jameysharp mentioned in https://github.com/bytecodealliance/sightglass/issues/202#issuecomment-1269162316 that it would be nice if measurement results were emitted as soon as they were collected. This could become possible by refactoring how the measurements are serialized by `serde`.

fitzgen:
Another thing we could do, just to provide feedback that something is happening and the process isn't just hanging, is to print a dot to `stderr` after each iteration (maybe also a newline after every 80 dots too).

abrown:
To take this one step further, I've been going back and forth about whether we should bring in a progress meter here. We know how many iterations need to be run so it's not hard to keep track of the percentage complete. I even took a look at crates.io at some options but they all seem snazzier (and bigger) than what is needed here. `cargo` has a built-in, single-file progress meter that could be used to. On the "con" side, this is just one more thing to maintain, which might break, etc., so I'm on the fence. @fitzgen (and others: @cfallin, @jameysharp), what do you think?

cfallin:
As someone who's spent a good amount of time staring at a sightglass invocation waiting for it to finish... some indication, even just "1% ... 2% ... 3% ... " textual output (adjust step size as desired), of how long I might be waiting would be very welcome! I've resorted to turning up `RUST_LOG` to see the individual invocations before to get some feedback.

fitzgen:
I would personally want to avoid any fancy progress bar, since it seems like it might introduce unnecessary noise into the system and seems a bit overkill.

abrown:
Ok, no progress bars. One thing I just looked up is whether we could detect an ANSI terminal and just rewrite the same line; [seems like it might work with `\r`](https://en.wikipedia.org/wiki/ANSI_escape_code#C0_control_codes) and there may be some [crates that can do this kind of thing](https://crates.io/search?q=ansi&sort=downloads).

PR #209, opened 2022-10-24
Display benchmark progress

abrown:
This change emits a string to `stderr` that indicates how far Sightglass has come in running its benchmarks:

```
Iterations completed: 42/100
```

The idea is to provide more immediate and clear feedback than `RUST_LOG` that Sightglass is doing work and the user can wait for it to finish. This turns out to be rather tricky to do due to the single-process and multi-process paradigms for running benchmarks; the solution is to communicate here with an internal environment variable "protocol."

The other tricky bit is the addition of ANSI terminal escape codes. This allows Sightglass to overwrite the same line and avoid filling up a screen with repeated `Iterations completed: ...` messages. This part of the PR could be safely removed without change to the counting functionality but I hope that the implementation is simple enough--and pessimistic enough--that it should not cause trouble down the road. I say "pessimistic" because the heuristic for determining if a terminal can support these "overwriting" terminal codes should just default to printing extra lines if anything looks amiss.

This PR is related to #203 and may close that issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1006,9 +1006,9 @@ dependencies = [
 
 [[package]]
 name = "once_cell"
-version = "1.13.1"
+version = "1.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "074864da206b4973b84eb91683020dbefd6a8c3f0f38e054d93954e891935e4e"
+checksum = "e82dad04139b71a90c080c8463fe0dc7902db5192d939bd0950f074d014339e1"
 
 [[package]]
 name = "openssl"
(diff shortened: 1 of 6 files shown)


abrown:
cc: @cfallin, @jlb6740, @jameysharp.

fitzgen (commented):
This is... a lot of code for printing some bits to `stderr` to show that something is actually happening. Especially when we start interacting with `RUST_LOG` and all that. I've also found tty detection to be fairly flaky in the past (e.g. doesn't work properly with terminals inside other tools, like Emacs). The communication between parent and child processes, etc.

Also: do the terminal escape codes work on windows too?

I feel like we can get 95% of the benefit with 1% of the code if we just do

```rust
eprintln!(".");
```

in the `execute_in_multiple_processes` function's loop after spawning each child process. This would give us a `.` after every process completes, which is 10 iterations by default. That seems to satisfy the "I can't tell if it is doing anything" use case just fine by me. If we really want more information, we could print `Process N/M complete.` or something like that. But we really don't need all the bells and whistles just to get some feedback...

fitzgen:
Opened #211 as an alternative approach

PR #215, opened 2022-12-16
Add task_clock as additional counter used with perf counters

jlb6740:
Currently all of sightglass's measuring units are tied processor events which is only useful in comparing performance where the cpu remains constant. This patch adds cpu time software counter as calculated by the kernel.


fitzgen (approved):

fitzgen:
The CI failure looks like we need to update our rustc in CI.

fitzgen:
https://github.com/bytecodealliance/sightglass/pull/218 should fix the CI issue.

abrown (approved):
LGTM with the CI fixed.

PR #217, opened 2022-12-30
Changing output slightly so that it's uniform

brianjjones:
Now it only prints that it successfully matched the image with at least 90% certainty. If it fails, it will print the error instead. This will allow the
benchmark to succeed on various platforms, which
may have slightly different confidence levels for
the result.


jlb6740:
Wow .. I was literally working on this patch when I saw the expected output had been updated to a confidence interval. 🥇

PR #218, opened 2023-01-03
Set nightly as the default toolchain in CI

fitzgen:
So that random direct invocations of `rustc` or `cargo` in our CI scripts get the right version.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -32,5 +32,7 @@ jobs:
       with:
         default: true
         toolchain: nightly
+    - name: Check rust version
+      run: rustc --version
     - name: Run benchmarks
       run: benchmarks/run-all.sh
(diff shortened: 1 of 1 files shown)


abrown:
Not sure this worked?

abrown:
Maybe something weird is going on with CI: I can see the task you linked to seems to have passed and the build environment that is emitted has the correct `1.68.0` version: https://github.com/bytecodealliance/sightglass/actions/runs/3875774546/jobs/6608773907#step:5:398. For the current failing task, though, it is back to `1.65.0` and it looks like the the `rustup default nightly` never happens: https://github.com/bytecodealliance/sightglass/actions/runs/3875774546/jobs/6608773907#step:5:398 (working) vs https://github.com/bytecodealliance/sightglass/actions/runs/3875774530/jobs/6608773021#step:3:17 (not working).

PR #220, opened 2023-01-18
New benchmark: RSA Blind Signatures

jameysharp:
This benchmark computes RSA Blind Signatures using the `blind-rsa-signatures` crate. Generating these signatures is a useful test of math on big integers.

The `blind-rsa-signatures` crate is licensed under the MIT license.


jameysharp:
Maybe this isn't slow enough as-is to be a good benchmark. Building a standalone version without the bench-api bits, this program runs in the 60-75ms range under Wasmtime. The same program compiled to native code takes 13-14ms though, so there's quite a bit of room for improvement and I'd like to capture that within the benchmark suite somehow.

abrown (approved):
LGTM! If you think of a way to increase the amount of work between `bench::start` and `bench::end` I will be happy to review that too! (More signatures? Larger signatures? Several iterations?).

PR #224, opened 2023-03-08
Add a regular expression benchmark

### LINKED ISSUE #216: Add a regex benchmark
I recently heard someone mention that there is a general assumption that "regex is slow in Wasm." This seems something that could be evaluated by sightglass so I started considering how one might structure such a benchmark: we would want to search for both simple and complex patterns, in small and large texts, with many matches and no matches, etc.

Does anyone know of such a benchmark that already exists? Or, alternately, is there a preference for how such a benchmark should be written? I was thinking it would be easiest to write some Rust code using the `regex` crate and compile it to WebAssembly and then compare this against its native execution. Any other opinions?

abrown:
This regular expression benchmark measures the time taken to match e-mails, URIs, and IP addresses in an input text. It is written in Rust and significantly adapted from the [regex-benchmark] project. The input text is a bit heavy (6.6MB) and the `benchmark.wasm` is as well (2.8MB); it may be worthwhile at some point to run some of these benchmark files through `wasm-opt`, e.g., to reduce the size.

It appears that one iteration of this benchmark takes around half a second for all three phases and the compilation phase is about an order of magnitude larger than the execution phase. I looked into creating a smaller version of the input text, but this would not significantly change the overall benchmark time (since it is concentrated in compilation) and, importantly, I could not figure out a way to have different `stdout` and `stderr` expectations for different workloads. With a smaller workload, the output _will_ be different and currently there is no way to express that.

[regex-benchmark]: https://github.com/mariomka/regex-benchmark

Closes #216.


abrown:
If people are interested, this workload might be one to add to the `default.suite`--it seems like interesting code to benchmark by default.

abrown:
cc: @fitzgen, @alexcrichton, @cfallin, @jameysharp, @elliottt, @jlb6740. Can someone take a look and +1?

fitzgen (approved):
LGTM!

jameysharp:
Awesome, I've wanted a regex benchmark to study in Sightglass. Thanks!

I've opened #226 with a bunch of experiments which somebody could try that should affect compile time, run time, or both, and in some cases may hit very different code generation paths in Cranelift.

PR #227, opened 2023-03-14
Fix for shootout-ackermann to read properly read input

jlb6740:


abrown:
@jlb6740, we should check in the updated `benchmark.wasm` here as well or eventually the CI will catch that there is a difference.

benchmarks/shootout-ackermann/benchmark.c
```diff
@@ -24,11 +24,12 @@ int read_int_from_file(char* path) {
     int fd = open(path, 0);
     assert(fd != -1);
 
-    ssize_t n = 0;
+    ssize_t m =0, n = 0;
```
abrown, line 27:
```suggestion
    ssize_t m = 0, n = 0;
```

abrown (approved):
See my comment about the `benchmark.wasm`, but this LGTM!

PR #230, opened 2023-03-17
Add some wasm simd using benchmarks

alexcrichton:
I added a hex encoder (not that interesting) using simd (maybe a little interesting?) which I've been using to drive some recent Cranelift improvements. Additionally I've updated the meshoptimizer benchmark to the latest version of the upstream repository and additionally updated it to use SIMD within the binary as well.


benchmarks/hex-simd/rust-benchmark/src/main.rs
```diff
@@ -0,0 +1,380 @@
...
+    {
+        if true {
+            return unsafe { hex_encode_simd128(src, dst) };
+        }
+    }
+
+    hex_encode_fallback(src, dst)
+}
+
+#[target_feature(enable = "avx2")]
```
abrown, line 44: Do we need to attribute this to another project?

abrown (approved):
LGTM!

benchmarks/hex-simd/rust-benchmark/src/main.rs
alexcrichton: The original source [comes from here](https://github.com/rust-lang/stdarch/blob/master/examples/hex.rs) of which I believe I'm the main substantive author of. Should I copy in the LICENSE file here to be safe though?

abrown: Heh, it's up to you I guess!

Issue #231, opened 2023-03-20
benchmark: failed to rebuild spidermonkey benchmark


abrown:
The "rebuild benchmarks" CI workflow found an issue rebuilding the SpiderMonkey benchmark: https://github.com/bytecodealliance/sightglass/actions/runs/4453084060/jobs/7821309649#step:4:5101. This CI workflow only rebuilds a subset of the benchmarks each CI run (for time's sake) to check that they can be rebuilt deterministically, so the failure will only reappear occasionally. cc: @cfallin

cfallin:
Hmm -- I took a look at this and was able to reproduce locally. I'm not sure where the `mach bootstrap` error is coming from though. The version of SpiderMonkey we have pinned is fairly outdated so it's possible that the toolchain it's trying to download no longer exists, and an update to latest SpiderMonkey would fix this. (I don't have the bandwidth to do this at the moment, but we can leave this open to track...)

william-stacken:
I was able to get around this by editing the Spidermonkey Dockerfile to clone the fork https://github.com/fastly/spidermonkey-wasi-embedding instead of https://github.com/tschneidereit/spidermonkey-wasi-embedding.
But for some reason the `runtime.cpp` gets stuck somewhere inside `JS_Init()` when I attempt to run the benchmark.

PR #232, opened 2023-03-22
Update wasm build scripts to wasi-sdk-19 and rust 1.67

jlb6740:
Patch updates wasm build scripts to use wasi-sdk-19 or rust 1.67 and also updates the wasm files built from those scripts. The only exception is spidermonkey which failed to build regardless of sdk version. This needs more investigation.


jlb6740:
Amazingly regex failed after upgrading to a newer version of rust due to a change in output:

```
     Running `target/debug/sightglass-cli benchmark --engine engines/wasmtime/libengine.so --raw --output-format csv -- benchmarks/regex/benchmark.wasm`
Error: Actual `stderr` does not match the expected `stderr`!
* Actual `stderr` is located at `stderr-1f1a1a7c163dcaf2-2151573-0.log`
* Expected `stderr` is located at `benchmarks/regex/stderr.expected`
Error: benchmark subprocess did not exit successfully
jlbirch@jlbirch-dev-20230321 22:48:41 ~/sightglass-jlb6740: more stderr-1f1a1a7c163dcaf2-2151573-0.log 
[regex] matching default.input
[regex] found 92 emails
[regex] found 5301 URIs
[regex] found 5 IPs
jlbirch@jlbirch-dev-20230321 22:48:49 ~/sightglass-jlb6740: more benchmarks/regex/stderr.expected 
[regex] matching ./default.input
[regex] found 92 emails
[regex] found 5301 URIs
[regex] found 5 IPs
```

So also updating the expected results file for regex.

abrown (approved):
Thanks for doing this! The build failure is for a different benchmark, spidermonkey, and already tracked in #231, so I think that is fine. I found it interesting that the majority of the benchmarks were reduced slightly in code size (low single digit percentage), one was reduced by half (!), and around three were increased by ten percent. If we ever need to figure out what language features impact code size we could dig more into this.

PR #236, opened 2023-05-12
Align ackermann output to its input

jlb6740:
In working on the patch #228 I found an issue with ackermann where I discovered there was a mismatch between the input and the output.


abrown (approved):

jlb6740:
Will merge this. It appears ackermann is not running properly. It looks like the working-dir flag is either not adequate or just not working properly. When both the wasm binary and the input are moved to the same directory from which the sightglass-cli is run then things work as expected, else the input files aren't read properly

Issue #239, opened 2023-05-17
Remove `+nightly`


abrown:
I don't think running Sightglass requires nightly any more (and has not for some time).

PR #241, opened 2023-06-06
Patch to remove nightly channel rquirement for cargo when building or…

jlb6740:
… running benchmarks

Can't remember now what problem prompted the creation of this issue #239 but this patch is to remove the nightly requirement for cargo building and running of benchmarks.


abrown (commented):
Searching around in the repository for "nightly", I see additional uses in `.github/workflows` and several other places. I think it would make sense to remove those as well since that way we will (1) avoid confusing users and (2) make sure in CI that `+nightly` is in fact not needed.

abrown:
I don't think we can actually merge this. Once I removed nightly from CI we get failures related to the precision crate missing something on Windows.

Issue #238, opened 2023-05-12
Benchmark shootout-ackermann is not working for Wasm. 


jlb6740:
Benchmark shootout-ackermann is not working for Wasm. Specifically the expected results do not match because the input files for m and n are not read. It appears the working-dir flag is either not working or not adequate.

jlb6740:
This may be an issue with the version of the wasi-sdk being used. Looking at strace output I do not even see an attempt to load the input file though I believe it is attempting to. But also, the assertation in the benchmark
https://github.com/bytecodealliance/sightglass/blob/f30862314d230b8986061a146881bee0b83a76b9/benchmarks/shootout-ackermann/benchmark.c#LL25C4-L25C4
 that checks against the failure of creating a FD at the open attempt is not working. Will not open a separate issue for that just yet.

PR #242, opened 2023-06-06
Add a run_all.sh deny list for benchmarks known not to work

jlb6740:
Benchmarks have breaks that go undetected. When they are found it is not always reasonable to expect that they will be fixed quickly. To prevent CI from failing from non issues, this patch adds a check for benchmarks known not to work .. i.e. https://github.com/bytecodealliance/sightglass/issues/238

### DIFF
--- a/benchmarks/run-all.sh
+++ b/benchmarks/run-all.sh
@@ -19,8 +19,22 @@ if [[ ! -f $ENGINE ]]; then
 fi
 
 # Benchmark each Wasm file.
+DENY_LIST=("shootout-ackermann")
 for BENCH_FILE in $(find $PROJECT_DIR/benchmarks -name benchmark.wasm); do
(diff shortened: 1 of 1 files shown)


fitzgen (commented):
It sounded like Andrew has a fix for Ackermann? If so, then maybe we can just wait on that? If not, then I think we should just remove the benchmark. It doesn't make sense to have benchmarks in the repo that we know aren't working and ignore elsewhere, and it will be easy to reinstate once fixed since it is in history and all that.

jlb6740:
@fitzgen .. Hi, yes @abrown and I spoke offline about this. This was meant to be a nice to have when Ackermann was not working and a fix wasn't obvious. This was a way to acknowledge that, instead of watching CI fail. But its definitely not needed for Ackermann now. I'll close. Thanks.

PR #243, opened 2023-06-08
Fix shootout-ackermann

### LINKED ISSUE #238: Benchmark shootout-ackermann is not working for Wasm.

abrown:
As reported in #238, the `shootout-ackermann` benchmark output did not match what was expected. While debugging this, several things became apparent:
- Relying on `0` as flags to `open` was the problem. `wasi-libc` defines `O_RDONLY` as `0x04000000`, e.g., so it may be that passing `0` is problematic for some reason. Once `O_RDONLY` is used, the benchmark works as expected.
- `assert()` is not aborting the program as one might expect. There are a couple of definitions in `wasi-libc` (e.g., [bottom-half] and [top-half]) that seem to do the right thing but I did not observe an abort when I passed in bad paths to `open()`.

[bottom-half]: https://github.com/WebAssembly/wasi-libc/blob/main/libc-bottom-half/headers/private/assert.h
[top-half]: https://github.com/WebAssembly/wasi-libc/blob/main/libc-top-half/musl/include/assert.h

Both of these issues seem wasi-libc related and might warrant some investigation there. This benchmark currently is building itself with wasi-sdk v19.0; by reverting this commit, here are the steps to replicate this issue:

```console
$ benchmarks/build.sh benchmarks/shootout-ackermann/
$ RUST_LOG=trace cargo run -- benchmark --engine=engines/wasmtime/libengine.so --processes 1 --iterations-per-process 1 benchmarks/shootout-ackermann/benchmark.wasm
```

In any case, using `O_RDONLY` seems to fix the issue so we can close
 #238 and continue the discussion in the wasi-libc repository.


jlb6740:
Great find Andrew!

abrown:
After some discussion today, it became clear that at least the `assert` issue is of our own doing: at some point we "undefined" `NDEBUG` which changes how `assert` works in C generally. https://github.com/bytecodealliance/sightglass/pull/244 fixes this.

PR #244, opened 2023-06-08
Removes define of NDEBUG to allow asserts to work

jlb6740:


fitzgen (approved):
Nice!

abrown:
The `benchmarks/run` CI task should be green if we rebase this on #243. But to get the `benchmarks/rebuild` task to work I think we may need to rebuild all of the benchmarks and check them back in to Git. Changing how we build the benchmarks will change the output WebAssembly and that is what that `benchmarks/rebuild` is doing--ensuring we build the same bytes in CI as we would locally.

abrown:
(or maybe ripping out `NDEBUG` doesn't affect the WebAssembly output?!?)

jlb6740:
Yes, if we modify the dockerfile used to build the benchmarks then all the wasm benchmarks that use that file should be checked in as well. I can submit a patch on top of this one that does that.

jlb6740:
Issue first discovered #238

PR #250, opened 2023-06-14
Add a `clean.sh` script

abrown:
It can be slightly annoying to clean up all the `.log` files Sightglass creates. Though it is quite easy to craft a `find` command to do this, it seemed efficient to record that in a script to avoid extra cognitive load.


abrown:
One thing that occurs to me looking at this is that we could move all the `.sh` scripts into a `scripts/` directory

fitzgen:
Can we actually add this to sightglass itself? I think we had plans to do this, or at least I remember talking/thinking about it at one point...

fitzgen:
Ah, yes we have https://github.com/bytecodealliance/sightglass/blob/17f71824cfbc982e7a4faa27e7bf7736315c5a3c/crates/cli/src/clean.rs#L10

abrown:
Heh, never mind then! Guess I forgot what parts lived in scripts and what parts in the CLI.

PR #256, opened 2023-06-16
Add benchmark name to `std{out|err}.expected`

abrown:
PR #251 introduces the ability to build multiple benchmarks within a single directory. Currently, Sightglass will look for the files `stdout.expected` and `stderr.expected` in the `.wasm` directory in order to check the benchmark output. These names would conflict in the multi-benchmark case, so this commit renames these files to `<benchmark name>.std{out|err}.expected`.


fitzgen (approved):
Thanks!

PR #260, opened 2023-07-06
Reorganize `shootout` benchmarks into a single directory

abrown:
Now that #251 and #256 make it possible for more than one benchmark to live in a single directory, this change moves all of the `shootout` artifacts into a single directory. The commits in this PR perform these steps atomically and can be reviewed sequentially. Even with the documentation I added in several places, this supposes a large decrease in the number of lines of code to maintain; the main benefit here is likely coming from using a single script to build all of the native benchmarks. Upon recompilation, the WebAssembly benchmarks have increased in size and I'm not exactly sure why.


fitzgen (approved):
Didn't look in detail, but moving them all to a shared directory makes sense to me.

jlb6740 (changes requested):
How do you run native benchmarks for shootout now? The READMEs need to be updated. From what I can tell this refactoring eliminates the cargo build system used for native that was just put in place?

jlb6740:
> How do you run native benchmarks for shootout now? The READMEs need to be updated. From what I can tell this refactoring eliminates the cargo build system used for native that was just put in place?

Note, when I tried native does build and seem to execute fine .. it's just are we ok with doing away with the Cargo build system and the update of the READMEs (in the native engine folder and the benchmark folder) where the former's instructions on how to build are now wrong and the later explicitly tells developers that the benchmark folder is required to supply a cargo based build if they want to support a native build.

jlb6740 (approved):
Cool. Looks good to me.

PR #264, opened 2023-07-14
Add's native support for meshoptimizer

jlb6740:


benchmarks/meshoptimizer/sightglass.native.patch
```diff
@@ -0,0 +1,111 @@
...
+
+ /**
+  * Call this function to indicate that recording should end. This call should be placed immediately
+@@ -17,7 +23,13 @@ void bench_start();
+  */
+ __attribute__((import_module("bench")))
+ __attribute__((import_name("end")))
++#ifdef __cplusplus
++extern "C" {
++#endif
```
abrown, line 104: Maybe we should just add these to the original header file?

abrown: And perhaps we just want to enclose the whole file in `extern "C"`?

benchmarks/meshoptimizer/build-native.sh
```diff
@@ -0,0 +1,13 @@
...
+# Build all of the shootout benchmarks as native shared libraries (Linux-only).
+#
+# Usage: ./build-native.sh
+
+(set -x; rm -r -f meshoptimizer);
+(set -x; git clone --recurse-submodules https://github.com/zeux/meshoptimizer.git);
+(set -x; cd meshoptimizer; git reset --hard f734fd572aed5bf76e84d9ed62ca6f4f6c47d84e; cd -);
+(set -x; cp sightglass.h sightglass.native.patch meshoptimizer/);
+(set -x; cd meshoptimizer; patch  -Np1 -i ./sightglass.native.patch; cd -);
+(set -x; cd meshoptimizer; CXX=/usr/bin/clang++ make codecbench-simd.so ; cd -);
```
abrown, line 12: You could add a `Dockerfile.native` file to run this script to match our new documentation for native-building. Then this can be built with the `build-native.sh` script.

benchmarks/meshoptimizer/build-native.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+# Build all of the shootout benchmarks as native shared libraries (Linux-only).
+#
+# Usage: ./build-native.sh
+
+(set -x; rm -r -f meshoptimizer);
```
abrown, line 7: If we're going to print every command we might as well just `set -x` once at the top.

benchmarks/meshoptimizer/build-native.sh
```diff
@@ -0,0 +1,13 @@
...
+(set -x; git clone --recurse-submodules https://github.com/zeux/meshoptimizer.git);
```
abrown, line 8: Will this newly-created directory be ignored by Git or should we add to a `.gitignore` in this directory?

benchmarks/meshoptimizer/sightglass.native.patch
jlb6740: > Maybe we should just add these to the original header file?

That was my original patch but unfortunately it doesn't work. The Wasm build for intgemm-simd failed. The exact error I can not remember .. but I think it was a linking error. Frankly, I think it is potentially an issue with the toolchain. I did assure myself that the issue went away when the extern C was removed. There was another failure as well of a different benchmark. All the shootout did pass.

jlb6740: Will patch for meshoptimizer instead of making changes seen by all benchmarks in order to not cause unexpected issues with any of those other benchmarks.

benchmarks/meshoptimizer/build-native.sh
jlb6740: Yes .. it shows. Will add.

abrown (approved):
LGTM!

PR #265, opened 2023-07-19
Fix native run script to grab any and all potential input files

jlb6740:


jlb6740:
We don't have a stipulation that says input file names need to be the same as the benchmark directory name and some benchmarks have multiple input files .. will remove the logic that checks only for input files that have the same name as the directory.

abrown:
Which benchmarks have `*.input` files that don't start with the benchmark name? I can change those just for consistency...

abrown (approved):
@jlb6740 showed me the ~10 other benchmarks that have files like `default.input` that need this change.

PR #268, opened 2023-08-30
Add `simd.suite`

abrown:
As discussed in a Cranelift meeting, it would be nice to know which benchmarks are SIMD users. This change adds a `simd.suite` file that lists the files that contain SIMD instructions. This can be run like:

```console
$ cargo run -- benchmark --engine .../libengine.so benchmarks/simd.suite
```

There are a couple ways this could be improved, but this may be fine as-is for now:
- `simd.suite` could get out of date; someone could create a CI check to keep it in-sync with the added files
- someone could really check that the measured section of each benchmark is actually a heavy SIMD user by collecting the instruction mix (e.g., `sde`)

### DIFF
--- a/benchmarks/simd.suite
+++ b/benchmarks/simd.suite
@@ -1,0 +1,89 @@
+# These benchmarks use WebAssembly SIMD instructions. Inclusion in this list
+# does not guarantee that the benchmark is a SIMD-centric workload in the
+# measured section, only that SIMD instructions are present.
+#
+# These files were gathered by:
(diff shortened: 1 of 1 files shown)


abrown:
cc: @alexcrichton, @fitzgen, @jameysharp, @cfallin

fitzgen (approved):

abrown:
I'm going to merge this in hopes that recent Wasmtime changes have fixed the build issues.

PR #270, opened 2024-04-24
Add a pure Wasm image classification benchmark

adambratschikaye:
Add a benchmark which performs image classification in pure Wasm using tract and ONNX. This can be useful to track the performance of float heavy computations.


adambratschikaye:
I think the rebuild failure is due to libsodium now requiring zig 0.12, but the Dockerfile uses zig 0.10.1.

adambratschikaye:
Rebuild failure should be fixed in https://github.com/bytecodealliance/sightglass/pull/271

benchmarks/tract-onnx-image-classification/setup.sh
```diff
@@ -0,0 +1,19 @@
...
+set -euxo pipefail
+
+MODEL_URL="https://www.github.com/onnx/models/raw/main/validated/vision/classification/mobilenet/model/mobilenetv2-7.onnx?download=:"
+MODEL_FILE="assets/mobilenetv2-7.onnx"
+
+if [ ! -f $MODEL_FILE ]; then
+	echo "Downloading model to $MODEL_FILE"
+
+	if which wget >/dev/null ; then
+	wget $MODEL_URL -O $MODEL_FILE
```
jlb6740, line 12: An assets directory needs to be created first.

jlb6740:
@adambratschikaye I was not familiar with tract but was introduced by your issue here: https://github.com/bytecodealliance/wasmtime/pull/8313, so thanks for submitting. I think getting this in now is great but ultimately maybe this directory could be refactored to include multiple inferencing scenarios.

benchmarks/tract-onnx-image-classification/setup.sh
adambratschikaye: Oops, thanks I'll fix that.

jlb6740:
@adambratschikaye Looks like you were right. Not sure the corruption, but it was using an incorrect cached version of the wasmtime-engine. I removed the cached image for macos  and all tests past now.

jlb6740 (approved):

PR #271, opened 2024-04-24
Bump Zig version for libsodium

adambratschikaye:
The current stable libsodium requires zig 0.11.0 to build. A couple of additional benchmarks have been introduced with the new stable release.


abrown (approved):
Thanks for this change!

Issue #267, opened 2023-08-03
image-classification benchmark failed


jlb6740:
root@4722daa600d1:/sightglass# /sightglass/target/release/sightglass-cli benchmark --processes=3 --engine /sightglass/engines/wasmtime/libengine.so --raw --output-format csv --output-file /sightglass/benchmarks//results//image-classification_wasm_results.csv -- /sightglass/benchmarks/image-classification/image-classification-benchmark.wasm
failed to instantiate

Caused by:
    unknown import: `wasi_ephemeral_nn::load` has not been defined
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `-1`,
 right: `0`', /sightglass/crates/recorder/src/bench_api.rs:221:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: benchmark subprocess did not exit successfully

rahulchaphalkar:
This error was due to not using `--engine-flags="--wasi nn"`.
I've opened a PR to make setup of environment of the benchmark simpler.

PR #273, opened 2024-04-30
Update image-classification benchmark setup

rahulchaphalkar:
Addresses https://github.com/bytecodealliance/sightglass/issues/267
Combines 2-step setup into 1-step, now only `source ./benchmarks/image-classification/setup.sh` is needed.
Updated Readme to reflect this.
@jlb6740

### DIFF
--- a/benchmarks/image-classification/README.md
+++ b/benchmarks/image-classification/README.md
@@ -4,9 +4,8 @@ This benchmark uses [wasi-nn](https://github.com/WebAssembly/wasi-nn), a propose
 learning (ML). It uses the OpenVINO backend to identify the subject of the test image, and measures the time required for a single inference to run.
 
 ## Steps to run this benchmark from the top sightglass directory
-- `./benchmarks/image-classification/setup.sh`
-- `source benchmarks/image-classification/openvino/setupvars.sh`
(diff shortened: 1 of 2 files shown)


jlb6740:
@rahulchaphalkar When I tried to run this, I got this error:

"[setupvars.sh] ERROR: Unsupported Python version. Please install one of Python 3.6 - 3.9 (64-bit) from https://www.python.org/downloads/"

This is coming from the OpenVino download where I usually just go in and comment out the lines myself. But it got me thinking if this is the latest version of OpenVino that will work here?

rahulchaphalkar:
> But it got me thinking if this is the latest version of OpenVino that will work here?

The openvino version here is old, 2022.x, as is the wasi-nn bindings version, 0.2.1. I haven't really updated anything in this PR, just tried to smooth over some setup script issues with openvino.

> "[setupvars.sh] ERROR: Unsupported Python version. Please install one of Python 3.6 - 3.9

As you pointed out this seems to be coming from setupvars.sh script from openvino. I think this should pop up irrespective of the changes in this PR right? Seems like it just needs an updated python to set up libopenvino.so path and stuff.

Issue #275, opened 2024-08-13
Missing Expected Output Files


d-sonuga:
The benchmarks in `benchmarks/libsodium` are missing their expected stdout and stderr files.

jlb6740:
I believe the checker here uses the basename of the wasmfile as the basis for the filename used for stderr. This probably should change to something more generic so that we don't have a bunch of redundant stderr.expected files. In this case all benchmarks are expected to return '0' to stdout. Will confirm and work on the patch to make a default.stderr.expected and default.stdout.expected when we don't care for the distinction.

PR #278, opened 2024-08-16
Add search for a default expected file and include for libsodium

jlb6740:
In some directories (at least libsodium) we have the same output expected for all benchmarks in the folder. Instead of creating 10s or even 100s of benchmark.std***. expected files we can have a default expected file that is used as a fall back if it exist. This patch implements this and also adds a expected files for libsodium to address #275.


jlb6740:
Note also, this is done in WasmScore and there has been no issue.

crates/cli/src/benchmark.rs
```diff
@@ -244,8 +244,14 @@ impl BenchmarkCommand {
             .with_context(|| "expected the benchmark file to have an extension")?
             .to_str()
             .with_context(|| "expected the benchmark file to have a printable name")?;
-        let stdout_expected = wasm_file_dir.join(format!("{}.stdout.expected", benchmark_name));
-        let stderr_expected = wasm_file_dir.join(format!("{}.stderr.expected", benchmark_name));
+        let mut stdout_expected = wasm_file_dir.join(format!("{}.stdout.expected", benchmark_name));
+        if !stdout_expected.exists() {
+            stdout_expected = wasm_file_dir.join(format!("{}.stdout.expected", "default"));
```
abrown, line 249:
```suggestion
            stdout_expected = wasm_file_dir.join("default.stdout.expected");
```

crates/cli/src/benchmark.rs
```diff
@@ -244,8 +244,14 @@ impl BenchmarkCommand {
...
+        }
+        let mut stderr_expected = wasm_file_dir.join(format!("{}.stderr.expected", benchmark_name));
+        if !stderr_expected.exists() {
+            stderr_expected = wasm_file_dir.join(format!("{}.stderr.expected", "default"));
```
abrown, line 253:
```suggestion
            stderr_expected = wasm_file_dir.join("default.stderr.expected");
```

abrown (commented):
From the CI run failure, is this default the right expectation for all the libsodium tests?

crates/cli/src/benchmark.rs
jlb6740: @abrown .. the code here is fine but I was under the impression that all the libsodium benchmarks returned 0 but some actually don't write anything to the log. Below is a cat of some of the logs when running all benchmarks in the suite:

```
stdout-d3b16d56d6c3a977-691414-2.log
::::::::::::::
0
::::::::::::::
stdout-d665976133d9c65-691414-0.log
::::::::::::::
0
::::::::::::::
stdout-d665976133d9c65-691414-1.log
::::::::::::::
0
::::::::::::::
stdout-d665976133d9c65-691414-2.log
::::::::::::::
0
::::::::::::::
stdout-d6c7b658a5a51c0a-691414-0.log
::::::::::::::
::::::::::::::
stdout-d6c7b658a5a51c0a-691414-1.log
```

To avoid modifications, I can just add empty expected output files for the ones that return nothing. But before I do that I'll look to see if it makes sense to return 0 so that all files share the same default.stdout.expected and default.stderr.expected files.

jlb6740: @abrown Resolved with the latest push to include a couple of non-default expected files.

abrown: It's good that you resolved the CI issues but that is not what this comment was originally about: there's no reason to generate a string with `format!` when we already know what the string value should be.

abrown (approved):

PR #281, opened 2024-12-19
Allow benchmarking just one particular compilation/instantiation/execution phase

fitzgen:
This avoids recompiling the Wasm module many times when we are only interested
in benchmarking instantiation or execution. Note that it is still compiled once
per process, but you can compile it exactly once if you do `--processes 1`.

cc @alexcrichton

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 members = [
     "crates/analysis",
     "crates/api",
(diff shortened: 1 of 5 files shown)


fitzgen:
Example usage that only benchmarks execution, not compilation or instantiation, and only compiles the Wasm module once:

```
$ sightglass-cli -e path/to/libengine.so --benchmark-phase execution --processes 1 --iterations-per-process 30
```

fitzgen:
Needs https://github.com/bytecodealliance/wasmtime/pull/9877 to land before CI will pass.

abrown (approved):

PR #283, opened 2025-04-09
Fix machine fingerprint byte size output

posborne:
Stumbled into this while exploring the cli options, was outputting that my machine had multiple TiB of RAM.  The version bump isn't necessary but I noticed it was there while looking at the docs.


abrown (approved):

abrown:
My guess is we should rebase this on #284 once that merges?

posborne:
Rebased

PR #284, opened 2025-04-14
Grab Bag of CI Enhancements

posborne:
This first set of changes includes the following:
- Removes outdated github actions recipes that are no longer functioning.
- Updates the benchmark flow to complete much more quickly by:
  1. Don't try to run 10 processes concurrently; this seems to be what previously caused some benchmarks to run for hours.
  2. Just do 1 iteration (for PR flow); nightly runs will be added later to get more complete results.
  3. Partition the test suite to run in 8 parallel jobs.
- Build wasmtime main and compare with version 25 (chosen arbitrarily) for test runs.  I suspect this will be a discussion point but this was somewhere to start.  Matrixing in other versions is very straightforward.

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -10,33 +10,81 @@ on:
 env:
   CARGO_TERM_COLOR: always
   RUST_LOG: info
+  PARTITIONS: 8
 
 jobs:
-  rebuild:
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v2
-    - name: Install nightly
-      uses: actions-rs/toolchain@v1
-      with:
-        toolchain: nightly
-    - name: Rebuild benchmarks
-      run: benchmarks/build-all.sh 5
+  # rebuild:
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #   - uses: actions/checkout@v4
+  #   - run: rustup update
+  #   - name: Rebuild benchmarks
+  #     run: benchmarks/build-all.sh
 
-  run:
+  build-wasmtime:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # for now, limit this a bit
+        revision: ["main", "v25.0.0"]
+    env:
+      REVISION: ${{ matrix.revision }}
     steps:
-    - uses: actions/checkout@v2
-    - name: Install nightly
-      uses: actions-rs/toolchain@v1
+    - uses: actions/checkout@v4
+    - run: rustup update
+    - name: Build Engine
+      run: |
+        pushd engines/wasmtime
+        rustc build.rs
+        ./build
+        popd
+    - name: Upload Wasmtime Shared Library
+      uses: actions/upload-artifact@v4
       with:
-        toolchain: nightly
-    - name: Run benchmarks
-      run: benchmarks/run-all.sh
+        name: wasmtime-${{ matrix.revision }}
+        path: engines/wasmtime/libengine.so
+
+  benchmark:
+    runs-on: ubuntu-latest
+    needs: build-wasmtime
+    strategy:
+      matrix:
+        partition: ["00", "01", "02", "03", "04", "05", "06", "07"]
+    steps:
+    - uses: actions/checkout@v4
+    - name: Downloads All Artifacts (root)
+      uses: actions/download-artifact@v4
+      with:
+        path: artifacts
+    - run: ls -al . artifacts/
+    - run: rustup update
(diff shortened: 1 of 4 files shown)


.github/workflows/benchmarks.yml
```diff
@@ -10,33 +10,81 @@ on:
...
+    - name: Partition Benchmark Suite
+      run: |
+        ceil_div() { echo "($1 + $2 - 1) / $2" | bc; }
+        tmpdir=$(mktemp -d)
+        grep -v '^\#' benchmarks/all.suite > $tmpdir/all.suite
+        line_count=$(wc -l "$tmpdir/all.suite" | awk '{print $1}')
+        lines_per_chunk=$(ceil_div "$line_count" "$PARTITIONS")
+        split -d -l "$lines_per_chunk" "$tmpdir/all.suite" "${tmpdir}/all.split."
+        cp "$tmpdir/all.split.${{ matrix.partition }}" benchmarks/split.${{ matrix.partition }}.suite
+        rm -rf "$tmpdir"
```
abrown, line 71: Seems like this parallelization logic could go in a script to be able to reproduce locally?

.github/workflows/benchmarks.yml
```diff
@@ -10,33 +10,81 @@ on:
...
+      run: cat "benchmarks/split.${{ matrix.partition }}.suite"
+
+    - name: Benchmark Partition ${{ matrix.partition }}
+      run: |
+        cargo run benchmark \
+        --engine=artifacts/wasmtime-main/libengine.so \
+        --engine=artifacts/wasmtime-v25.0.0/libengine.so \
+        --processes=4 \
+        --iterations-per-process=1 \
+        -- "benchmarks/split.${{ matrix.partition }}.suite"
```
abrown, line 83: The `run.sh` script chooses slightly different parameters here. I'm not sure if it matters which parameters we use but I think making it easy to reproduce locally would be of some benefit.

abrown (commented):
Thanks for the PR! I had some comments below, but am interested to hear where you're headed &mdash; any new use cases for Sightglass?

.github/workflows/benchmarks.yml
posborne: Yeah, part of the issue with CI is conflicting defaults between `run-all.sh` and what we want for different CI flows.  These could be made configurable via more env vars, etc. on the `.sh` script; my reasoning for putting the invocation inline had to do with the somewhat coupled steps taken to do the parallel test execution (creating the splits in a previous job and then running the partition based on matrixing) that would probably muddy up the script as none of that job distribution logic would make sense outside of a CI context.

We also build wasmtime(s) in an earlier stage; run-all.sh is convenient but overloaded for achieving better CI performance.

For processes/iteration-per-process, I went with a combo which gives us about the same amount of data but at reduced execution time.  What I plan to implement in later changes is a workflow that can run on a scheduled basis (nightly or otherwise) that will be longer-running and generate data which acts as a better reference, possibly working towards generating something publishable to gh-pages for easy reference; if results are consistent enough across runs (not optimistic on gh runners providing that), that could be extended to create something akin to https://arewefastyet.com.

.github/workflows/benchmarks.yml
posborne: Yeah, I can extract this into a script.  Apart from testing for CI changes, I don't think it would provide anything meaningful locally, but it could be useful for developers who need to make changes.

.github/workflows/benchmarks.yml
```diff
@@ -10,33 +10,81 @@ on:
...
 
 jobs:
-  rebuild:
+  # rebuild:
+  #   runs-on: ubuntu-latest
+  #   steps:
+  #   - uses: actions/checkout@v4
+  #   - run: rustup update
+  #   - name: Rebuild benchmarks
+  #     run: benchmarks/build-all.sh
```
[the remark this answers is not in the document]
posborne: Ah, meant to revisit this so thank you for commenting.  I understand its inclusion now; I was a bit thrown off by the rebuild being performed but the results being discarded previously and wasn't sure what the intention was.

abrown (approved):
Looks good to me! Thanks for fixing some of those broken CI bits. I'll leave this open for a day or so in case you were planning to re-organize your commits (I see a `fixup!` in there) but, honestly, I will squash-merge this anyways so it's up to you.

posborne:
@abrown I'm happy enough with the changes here for now, so I would say squash-and-merge away is good with me whenever.  I'm sure we'll revisit some of the details here in future changes, but this at least gets CI going again in some form.

PR #285, opened 2025-04-17
Time Measure & Support for Multiple Measures

posborne:
You can now, for example, do the following to capture timing, cycles, and perf-counters data on a single run...

```
cargo run -- benchmark \
    --engine engines/wasmtime/libengine.dylib \
    --raw -o result.json \
    -m time -m cycles -m perf-counters \
    -- benchmarks/all.suite
```


abrown (approved):
LGTM!

PR #287, opened 2025-05-07
bump spidermonkey benchmark

chenyan2002:
The current `Dockerfile` doesn't pin the version of `spidermonkey-wasi-embedding`, so `docker build` cannot run, due to the mismatch of versions. This PR also bumps spidermonkey to the latest release.


abrown (approved):
This all looks fine to me; @cfallin, any issues with the SpiderMonkey part of this?

cfallin:
I haven't followed the evolution of the SpiderMonkey APIs here in detail, but in principle the changes should be fine as long as the benchmark continues to run -- that looks like it's tested in CI so 👍

PR #291, opened 2025-11-05
Fix CI Issues

posborne:
From #290 there were some CI failures unrelated to that change; attempting to fix and simplify things a little bit.

- Drop nightly toolchain -- I'm not aware of a reason that is required at this point.
- Pin libsodium to the single tagged release that exists rather than chasing stable (which had broken when using zig 0.12).
- Bump the wasmtime REVISION; v25 and the older commit no longer compile with stable it appeared.

I'll add commits if CI fails with additional bits to address.

### DIFF
--- a/.github/workflows/sightglass.yml
+++ b/.github/workflows/sightglass.yml
@@ -53,3 +53,5 @@ jobs:
 
     - name: Test all
       run: cargo test --verbose --all
+      env:
+        RUST_BACKTRACE: 1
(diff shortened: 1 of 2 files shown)


posborne:
https://github.com/jedisct1/rust-precision/pull/13 has the upstreaming request for the precision crate.  I'm still not sure why the linking issue on windows with the C version arose, but making the suggested change there seemed worthwhile and easier than debugging the linker issue.

PR #292, opened 2025-11-18
Add four benchmarks from JetStream 3: richards, quicksort, hashset, gcc-loops.

cfallin:
These benchmarks are included in [JetStream], which is nominally a browser JavaScript benchmark but which in recent times has started to include JS+Wasm benchmarks. I picked several of the Wasm benchmarks that are entirely or almost entirely in Wasm, with only the barest of JS wrappers, and tweaked them to build on WASI instead. Then I added the Sightglass-specific benchmark begin/end hook calls.

The original source files for these benchmarks remain under their original licenses. A license exception for inclusion of files under licenses other than `Apache-2.0 WITH LLVM-exception` (the BA's default license) was granted by the Bytecode Alliance board [here], and all files added in this PR are under `benchmarks/` and thus fit within that exception.

[JetStream]: https://github.com/webkit/jetstream
[here]: https://github.com/bytecodealliance/governance/pull/137


fitzgen:
r=me with CI passing

fitzgen:
Re-running

PR #295, opened 2025-12-16
Add `report` subcommand to generate html report

posborne:
This is a port of the viz.py proof-of-concept (#286).  The workflow for generate these reports is to:

1. Capture raw data to JSON/CSV.
2. Generate a report using 2 or more input data sets with one engine/flags designated as the baseline (by default the first engine found in the provided data files).

The report itself tries to provide clear, concise representation of the data that is statistically significant in the overview table with raw results data in detail sections which, along with statistics, can give an indication of how noisy and trustworthy differences in measured results for a run might be.

The report is intended to be shared; it does pull in a couple dependencies for graph generation from a CDN but is otherewise standalone.

---

Here's a report that was generated from this changeset showing:
* Baseline: wasmtime v32 with no flags
* wasmtime v39.0.1 with no flags
* wasmtime v39.0.1 with epoch-interruption=y

Example Report:

[report.html](https://github.com/user-attachments/files/24180034/report.html)

Screenshot of ^^^ to pique interest:

image

### DIFF
--- a/crates/analysis/src/keys.rs
+++ b/crates/analysis/src/keys.rs
@@ -140,7 +140,6 @@ mod tests {
             phase: Phase::Compilation,
             event: "cycles".into(),
             count: 42,
-            engine_flags: None,
         }));
     }
 }
--- a/crates/analysis/src/summarize.rs
+++ b/crates/analysis/src/summarize.rs
@@ -165,7 +165,6 @@ mod tests {
                 phase: Phase::Compilation,
                 event: "cycles".into(),
                 count,
-                engine_flags: None,
             }
         }
 
@@ -200,7 +199,6 @@ mod tests {
                 phase,
                 event: "cycles".into(),
                 count,
-                engine_flags: None,
             }
         }
         let measurements = vec![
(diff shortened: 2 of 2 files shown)


fitzgen (approved):
Looks great, thanks!

Not something we need to do *now* but it would be cool to integrate the HTML reports into our existing output format machinery, rather than (or in addition to) having it be a separate CLI subcommand, so we could just do something like

```shell-command
$ sightglass benchmark --output-format html ...
```

Also, it would be nice to show the confidence interval as a percentage (in addition or instead of the absolute value) since, at least for me personally, the absolute values are much less useful at a glance than the percentage. Maybe the "Performance" column should include a `+/- 1.23%` or something? Lots of ways we could display this, don't care too much about the details, more just that we try to make things as minimally noisy and easy to read at a glance as possible.

Thanks again!

posborne:
Thanks for the review @fitzgen; I created issues to track the follow-up work.  If you're OK with it, I would like to merge this as-is and visit those items later.

PR #302, opened 2025-12-19
Add TinyGo JSON/Regex serialization benchmarks

posborne:
- Create Dockerfile.tinygo template for TinyGo benchmarks
- Add tinygo-json benchmark with encoding/json
- Add tinygo-regex benchmark using standard library regexp

Sample size of one, but in some quick comparison the regex with tinygo is fairly slow taking ~32x which is somewhat surprising.  CC @dgryski to review to ensure the workloads here are useful in representing some proxy of real-world workloads.

Given how much slower tinygo appears to be here it may make sense to truncate the input file to reduce the per-iteration time of the regex benchmark.

---

As with the previous round of benchmarks, portions of this PR were generated using an LLM though reworked a fair bit.  I opted to use a single tinygo/ benchmarks subdirectory as it was easier to achieve reuse in that way that make it simpler to add other tinygo benchmarks as we see fit.


benchmarks/tinygo/build.sh
```diff
@@ -0,0 +1,34 @@
...
+
+    # Skip if not a benchmark directory (no main.go)
+    [ ! -f "$dir/main.go" ] && continue
+
+    echo "Building TinyGo $benchmark benchmark..."
+
+    cd "$dir"
+
+    # Build with TinyGo
+    "$TINYGO" build -o "$OUTPUT_DIR/tinygo-$benchmark.wasm" -target=wasi -opt=2 -gc=leaking .
```
fitzgen: Can you add a comment here documenting why `-gc=leaking` is used here?

fitzgen (approved):

PR #306, opened 2025-12-31
Report the confidence interval as 0.0 when there is no variance

fitzgen:
This is both better UX (no single error prevents viewing the rest of the non-error results) and more accurate (if there is no variance then we know the exact delta).


posborne (approved):
Makes sense to me.

PR #307, opened 2026-04-25
Add a `splay` Wasm GC benchmark

fitzgen:
This is a port of the `splay` benchmark from JetStream 3, which is itself
adapted from Octane 2. It stresses the performance of the garbage collector.


fitzgen:
FWIW, the "benchmarks" CI job is failing on `main`; not related to this PR.

benchmarks/default.suite
```diff
@@ -5,4 +5,4 @@
 
 bz2/benchmark.wasm
 pulldown-cmark/benchmark.wasm
-spidermonkey/benchmark.wasm
+spidermonkey/spidermonkey-markdown.wasm
```
posborne, line 8: Ah, I assume this was a goof from my previous set of patches.  I'm assuming we don't want splay here?  Probably just a fix.

posborne (approved):

benchmarks/default.suite
fitzgen: I think we probably want splay in the default suite at some point, but I kind of want to do that as part of a larger clean up pass of all of our benchmarks and defaults

PR #313, opened 2026-06-05
feat: add callgrind measure

ricochet:
Add CallgrindMeasure and a concept of process wrapping to benchmarks so that
each CallgrindMeasure is run under valgrind.

Additionally adds a gh workflow that runs the new callgrind measure nightly and on pushes to main.


ricochet:
This is what the summary output looks like: https://github.com/ricochet/sightglass/actions/runs/27025357300

fitzgen:
Closing in favor of #314

PR #314, opened 2026-06-05
Add a `callgrind` measure

### LINKED ISSUE #312: Add callgrind measure
It would be nice to have (nearly)deterministic measures that we could run in places like GitHub runners without introducing too much variance and noise between runs.

A couple references:
- rustc-perf takes the zero-client-request approach. It runs the entire rustc invocation under `valgrind --tool=callgrind --cache-sim=no --branch-sim=no --callgrind-out-file=clgout, then post-processes` and then post-processes. That works fine for rustc but in Sightglass we have separate phases (compilation + instantiation + N×execution).
- [crabgrind](https://crates.io/crates/crabgrind) C shim via cc + bindgen. drags in libclang + C compiler as Linux build deps...
- [valgrind-requests](https://docs.rs/valgrind-requests/latest/valgrind_requests/) inline asm (part of gungraun project) x86_64/aarch64/arm/riscv64/s390x/ppc on Linux + BSDs
- [gungraun](https://docs.rs/gungraun/latest/gungraun/) full harness so not what we would use. Interesting project though and might be worth referencing.

`cm-online-stats` seems like a great bench to start with for per-instruction attribution of cross-component call overhead. The Execution-phase dump will show exactly how many instructions each call spends in component trampolines/adapters and resource-table operations in the host runtime.

fitzgen:
This commit adds a new `callgrind` measure. It must always be run inside a child process that is running under Valgrind's Callgrind tool. It uses the `valgrind-requests` crate to communicate with Valgrind and record data from the simulated caches and branch predictor.

Running under Callgrind is much slower than running natively, but also is much less noisy. Therefore we adjust the default numbers of processes and iterations per process accordingly.

Fixes https://github.com/bytecodealliance/sightglass/issues/312

### DIFF
--- a/crates/cli/src/benchmark.rs
+++ b/crates/cli/src/benchmark.rs
@@ -238,6 +238,9 @@ pub struct BenchmarkCommand {
     engine_flags: Option<String>,
 
     /// How many processes should we use for each Wasm benchmark?
+    ///
+    /// Defaults to `10`, unless using the `callgrind` measure, in which case the
+    /// default is `3`.
     #[structopt(long = "processes", value_name = "PROCESSES")]
     processes: Option<usize>,
 
@@ -250,6 +253,9 @@ pub struct BenchmarkCommand {
     names: Option<Vec<String>>,
 
     /// How many times should we run a benchmark in a single process?
+    ///
+    /// Defaults to `10`, unless using the `callgrind` measure, in which case the
+    /// default is `1`.
     #[structopt(
         long = "iterations-per-process",
         value_name = "NUMBER_OF_ITERATIONS_PER_PROCESS"
--- a/crates/cli/src/benchmark.rs
+++ b/crates/cli/src/benchmark.rs
@@ -11,10 +11,190 @@ use std::{
     fs,
     io::{self, BufWriter, Write},
     path::{Path, PathBuf},
-    process::Command,
-    process::Stdio,
+    process::{Command, Stdio},
 };
 use structopt::StructOpt;
+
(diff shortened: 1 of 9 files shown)


crates/cli/src/benchmark.rs
```diff
@@ -58,8 +238,8 @@ pub struct BenchmarkCommand {
     engine_flags: Option<String>,
 
     /// How many processes should we use for each Wasm benchmark?
```
posborne, line 240: Might be nice to document the default here so it shows up in help output again as a hint to users looking to modify things.  Same applies for `iterations-per-process`.

posborne (approved):
Changes look good to me; I was able to do a couple runs with callgrind to confirm.

The time to run is definitely astronomically slow; if the data is reliable with a smaller sample, we may want to see about coming up some smaller inputs.  Possibly a different default suite when targeting callgrind compared with the default.  That's, of course, secondary to getting results that can be trusted but there's some balance point in there.

fitzgen:
> The time to run is definitely astronomically slow; if the data is reliable with a smaller sample, we may want to see about coming up some smaller inputs. Possibly a different default suite when targeting callgrind compared with the default. That's, of course, secondary to getting results that can be trusted but there's some balance point in there.

For sure, I am planning on doing a pass over the benchmarks to get them all running roughly the same amount of instructions per execution iteration when I have a chance. Probably won't be for every single one, but the ones that are easy enough to do that, I will.

FWIW, I will also be making a PR for the PCA stuff soon. Have it working locally, just need to do some final tweaks.

PR #315, opened 2026-06-08
Add a `pca-metrics` command

fitzgen:
This records various static and dynamic metrics for use in principal component analysis.

Static metrics include:

* Various entity counts: number of functions, number of instructions, number of data segments, etc...
* Static: instruction mix: ratio of integer vs float vs memory load vs etc... instructions

There are two categories of dynamic metrics:

1. Dynamic Wasm instruction mix: This is recorded via instrumenting the benchmark with counters for each instruction category and incrementing an instruction's associated counter as the program runs (with an optimization to batch these increments up once per basic block).

2. Callgrind-based metrics: The benchmark is run through callgrind with a fixed cache and branch predictor simulation configuration (so the metrics are independent of the host machine's microarchitecture) and then we report the cache read/write miss ratio, branch prediction miss ratio, number of Wasm instructions executed per native instruction, etc...


fitzgen:
@posborne gentle review ping

posborne (approved):
Sorry for the delay in review @fitzgen; couple false starts before I found the time to work through reading the full set of changes.  Looks good to me, I haven't had a chance to interact with the code yet, though I hope to a bit in the near(ish) future.

fitzgen:
Thanks @posborne !

PR #317, opened 2026-06-22
Add a script to run PCA and clustering of our benchmarks

### LINKED ISSUE #98: sightglass-next: add principal component analysis
To minimize the number of benchmarks needed to get a representative result, we could run PCA on the executed benchmarks. This would likely be implemented in R. There is a lot more discussion needed for this but this issue serves as a placeholder for that.

fitzgen:
The methodology is based on ["A Workload Characterization of the SPEC CPU2017 Benchmark Suite" by Limaye and
Adegbija](https://tosiron.com/papers/2018/SPEC2017_ISPASS18.pdf).

Each metric is standardized (centered to mean 0, scaled to unit variance) and PCA is run on the resulting correlation matrix so that metrics measured on different scales contribute comparably. Benchmarks are then clustered by the Euclidean distance between their principal-component scores, as in the paper.

Finally, we recommend a subset of the suite. Each cluster is represented by its cheapest member (the benchmark that executes the fewest dynamic wasm instructions). Sweeping the number of clusters traces a Pareto trade-off between clustering error (SSE) and the cost of running the subset (its total dynamic instructions); the knee of that curve is the Pareto-optimal cluster size.

Fixes https://github.com/bytecodealliance/sightglass/issues/98

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+*.svg
+
 **/*.rs.bk
 **/*~
 /target/
--- a/scripts/pca.R
+++ b/scripts/pca.R
@@ -1,0 +1,450 @@
+#!/usr/bin/Rscript --vanilla
+
+# Principal component analysis (PCA) and agglomerative hierarchical clustering
+# of Sightglass benchmark metrics.
+#
+# Before running the first time:
+#
+#     $ R
+#     > install.packages(c("FactoMineR", "factoextra", "svglite", "ggplot2"))
+#
+# Usage:
+#
+#     $ cargo run -- pca-metrics -o ./pca-metrics.csv -- benchmarks...
+#     $ ./scripts/pca.R ./pca-metrics.csv
+#
+# The methodology is based on "A Workload Characterization of the SPEC CPU2017
+# Benchmark Suite" by Limaye and Adegbija:
+#
+#     https://tosiron.com/papers/2018/SPEC2017_ISPASS18.pdf
+#
+# Each metric is standardized (centered to mean 0, scaled to unit variance) and
+# PCA is run on the resulting correlation matrix so that metrics measured on
+# different scales contribute comparably. Benchmarks are then clustered by the
+# Euclidean distance between their principal-component scores, as in the paper.
+#
+# Finally, we recommend a subset of the suite. Each cluster is represented by
+# its cheapest member (the benchmark that executes the fewest dynamic wasm
+# instructions). Sweeping the number of clusters traces a Pareto trade-off
+# between clustering error (SSE) and the cost of running the subset (its total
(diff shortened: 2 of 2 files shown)


cfallin (approved):
This looks "reasonable" to me, though of course I'm no data scientist (I just play one on GitHub Reviews apparently).

I think the real validation will come with the results. I know you've posted the plots that come out of this elsewhere -- it'd be great to see them here, together with some empirical validation of the clustering's representative behavior (i.e., changes that cause a swing in the full suite cause a swing for just the top-1-of-each-cluster, though possibly with different effect size since we are not computing weights here and benchmark-suite weighting is an artifact of ad-hoc benchmark collection anyway).

fitzgen:
> I think the real validation will come with the results. I know you've posted the plots that come out of this elsewhere -- it'd be great to see them here, together with some empirical validation of the clustering's representative behavior (i.e., changes that cause a swing in the full suite cause a swing for just the top-1-of-each-cluster, though possibly with different effect size since we are not computing weights here and benchmark-suite weighting is an artifact of ad-hoc benchmark collection anyway).

Yeah that's my plan, I just didn't want to gate this script landing on doing those experiments.

Roughly my next steps are:

* For the benchmarks that read `default.input`, try and right-size that input so they all take a reasonable amount of execution "time" (was going to go for number of Wasm instructions dynamically executed). Something long enough that the native results won't be too noisy and we potentially fill up our callgrind-simulated caches and get LL cache misses, but also short enough that the benchmarks still run relatively quickly. I'm thinking of targeting say ~5M Wasm instructions executed?
  * Will send a PR for this once done
* Optionally: add new benchmark candidates for programs/toolchains/whatever that we've wanted to have for a while (now is the time to do it, before running the "official" PCA analysis and/or archiving redundant benchmarks)
  * These would also be PRs
* Re-collect PCA metrics (this takes a few hours for the whole set of candidates, hopefully a little faster after the previous task)
* Run the PCA analysis, create graphs, etc... (fast, takes just a few seconds)
* Do the fake regression experiment and check that the measured regression(s) in the PCA-recommended subset of candidates reflects the measured regression(s) of the full set of candidates (that is, we shouldn't catch the regression(s) in the full set but not in the PCA-recommended subset)
* Present the results of the PCA analysis and fake-regression experiment in a Cranelift and/or Wasmtime meeting, along with a recommendation for a candidate subset to become The Official Sightglass Benchmark Suite(tm)
* Assuming we reach consensus on choosing a subset to become The Official Sightglass Benchmark Suite(tm), archive the benchmark candidates that are not in that suite
  * This would be another PR

Let me know if you have suggestions on how to improve that sequence.

cfallin:
That plan sounds great to me!

PR #318, opened 2026-06-24
Adjust workload sizes to ~100M Wasm instruction executed

fitzgen:
Running ~100M Wasm instructions takes ~.1 seconds, so this seems like a pretty reasonable number (100 execution iterations will take ~10 seconds (compilation adds more time ofc)).

Also make a few more benchmarks read their input from disk so it is easily tunable and also impossible for the optimizer to const-eval the whole benchmark away.


benchmarks/blake3-simd/benchmark.c
```diff
@@ -1,24 +1,84 @@
...
+    size_t len;
+} wasi_iovec_t;
+
+WASI_IMPORT("path_open")
+int wasi_path_open(int fd, int dirflags, const char *path, size_t path_len,
+                   int oflags, uint64_t rights_base, uint64_t rights_inheriting,
+                   int fdflags, int *opened_fd);
+
+WASI_IMPORT("fd_read")
+int wasi_fd_read(int fd, const wasi_iovec_t *iovs, size_t iovs_len, size_t *nread);
```
cfallin, line 27: A little unfortunate to bake in WASIp1 -- if we have refactor-energy here, maybe we could point it toward using wasi-sdk instead?

benchmarks/intgemm-simd/benchmark.cpp
```diff
@@ -12,13 +12,56 @@
...
 #include <cstdlib>
 #include <cstring>
 #include <iomanip>
 #include <iostream>
 #include <random>
 
+// This benchmark is built with Emscripten, whose libc filesystem does not reach
+// WASI preopened directories. So, like the splay and blake3-simd benchmarks, we
+// read the workload size from disk by calling the WASI `path_open`/`fd_read`
+// syscalls directly. This lets the workload be resized without recompiling.
```
cfallin, line 26: Likewise here -- use wasi-sdk instead?

benchmarks/intgemm-simd/benchmark.cpp
```diff
@@ -105,24 +148,29 @@ template <class Backend> void Print(std::vector<std::vector<double>> &stats, std
...
 
   RandomMatrices *matrices_end = (RandomMatrices*)matrices + sizeof(matrices) / sizeof(RandomMatrices);
 
   bench_start();
 
   BackendStats stats;
-  const int kSamples = 3;
-  std::cerr << "SSSE3 8bit, 3 samples..." << std::endl;
+  // One sample is enough: Sightglass counts executed instructions
+  // deterministically, so repeated timing samples would only inflate the count.
```
cfallin: This comment seems overly specialized to our current benchmarking harness plans ("Sightglass counts executed instructions deterministically"). The conclusion is actually correct since Sightglass has its own sampling loop, so benchmarks shouldn't need to do internal sampling; just thought it seemed incorrectly justified is all.

cfallin (approved):

posborne (approved):

benchmarks/blake3-simd/benchmark.c
fitzgen: I can look into this in a follow up

PR #319, opened 2026-06-24
New benchmark: a port of Richards to Kotlin/Wasm

fitzgen:
A Kotlin port of the classic Richards benchmark: a simulation of an operating system's scheduler. This version is a Wasm component produced via Kotlin/Wasm's `wasmWasi` target and `wasm-tools component new`.

Richards is an integer/pointer-heavy workload built around linked lists of "packets" routed between cooperating tasks (idle, worker, handler, device). It exercises object allocation, virtual dispatch, and pointer chasing, and here it also exercises the WasmGC, typed-function-references, and exception-handling proposals that Kotlin/Wasm relies on.


posborne (approved):
Looks good; definitely a good workload to exercise paths we aren't covering well today (gc/exception handlers).

fitzgen:
Thanks!

PR #320, opened 2026-06-25
Migrate from structopt to clap

posborne:
Structopt is long deprecated and unmaintained; it's features were rolled into clap's derive functionality, though with a few small differences.  I added some additional, basic test coverage and then migrated the cli to use clap.


fitzgen (approved):
Thanks! Long overdue

PR #321, opened 2026-06-25
Migrate the Emscripten benchmarks to `wasi-sdk`

fitzgen:
`noop`, `blake3-simd`, and `intgemm-simd` were the only benchmarks still built with Emscripten. They now build with `wasi-sdk`, matching the rest of the C/C++ benchmarks and removing the Emscripten dependency.

`noop` was straightforward since it doesn't do anything.

`blake3-simd` and `intgemm-simd` were trickier. Emscripten comes with shims for mapping x86 SSE intrinsics onto Wasm SIMD, but `wasi-sdk` doesn't do that, so this commit introduces `wasm_sse_compat.h` headers to fill that gap ourselves. Note that `blake3-simd` and `intgemm-simd` use almost non-overlapping sets of SIMD instructions, and one is C and the other C++, so I didn't bother trying to share a single `wasm_sse_compat.h` file between them.


cfallin (approved):
Thanks!

PR #322, opened 2026-06-25
Adjust `pca.R` to choose a cluster representative closest to 100M Wasm instructions executed

fitzgen:
Rather than the fewest Wasm instructions executed. 100M was chosen previously as a large enough number to avoid lots of noise and to fill our callgrind-simulated caches, but small enough to run individual iterations relatively quickly.


cfallin (approved):

cfallin:
Ah, oops, this repo doesn't have a test-gated merge like Wasmtime does -- I clicked 'merge' expecting it would do so after tests passed. If they don't, now, happy to review followups; sorry about that...

PR #326, opened 2026-07-09
A handful of changes to improve human-readable output and offer better UX

fitzgen:
Here is an example of the new output:

Screenshot 2026-07-09 at 1 13 16 PM


fitzgen:
(Thanks for all the reviews lately @posborne; almost done here!)

posborne (approved):
This is great, thanks for making these changes @fitzgen.  Definitely an improvement to usability.

I have considered some if we want to change the defaults further to default to capturing raw results and separating data collection from analysis even further.  Given how long it can often take to capture the data for a benchmark run, not having the raw results captured seems unfortunate.  In any case, these changes are good regardless and I believe the tooling for doing cli-based analysis continues to work using inputs from file.

No objections to landing; just tangential thoughts on UX for typical/optimal usage.

fitzgen:
> I have considered some if we want to change the defaults further to default to capturing raw results and separating data collection from analysis even further. Given how long it can often take to capture the data for a benchmark run, not having the raw results captured seems unfortunate. In any case, these changes are good regardless and I believe the tooling for doing cli-based analysis continues to work using inputs from file.

For sure, this is something I've thought about before as well. On the one hand, as you point out, it sucks to re-run benchmarks when you just want to analyze the results in a slightly different way. On the other hand, people want a single command that DWIM and "just works"; putting any friction on that path means they will just do a one-off test, probably with worse methodology and less representative benchmark programs.

I've been thinking that a best-of-both-worlds solution could be to do something similar to what `perf record` does with its `perf.data` files: always write the recorded data to `sightglass.csv`, but if there was already a `sightglass.csv`, then first move that to `sightglass.old.csv` (and accept that existing `sightglass.old.csv`s will be overwritten). This way we could both preserve raw results and have the DWIM behavior (printing human-readable effect size reports and all that) by default.

What do you think of this idea?

PR #328, opened 2026-07-22
Allow engine flags to be specified multiple times

fitzgen:
Change `--engine-flags` to accept multiple values and pair them with the engines as follows:

- A single engine with multiple flag sets is paired with each set

- Multiple engines with zero or one flag set pairs each engine with the flag set (if any)

- N engines and N flag sets are paired together by position: `engines[i]` with `flags[i]`

- All other combinations are errors.

This allows, for example, comparing the performance effects of different flags for the same engine, e.g. the DRC vs copying collector:

```
sightglass-cli benchmark \
    -e engines/wasmtime/libengine.dylib \
    --engine-flags "-Ccollector=copying" \
    --engine-flags "-Ccollector=drc" \
    -- benchmarks/splay/splay.wasm
```

### DIFF
--- a/benchmarks/Dockerfile.rust
+++ b/benchmarks/Dockerfile.rust
@@ -1,4 +1,4 @@
-FROM rust:1.87
+FROM rust:1.96
 RUN rustup target add wasm32-wasip1
 # Install protobuf compiler for rust-protobuf benchmark
 RUN apt-get update && apt-get install -y protobuf-compiler && rm -rf /var/lib/apt/lists/*
--- a/benchmarks/tract-onnx-image-classification/Dockerfile
+++ b/benchmarks/tract-onnx-image-classification/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:1.88
+FROM rust:1.96
 RUN rustup target add wasm32-wasip1
 WORKDIR /usr/src
 ADD rust-benchmark rust-benchmark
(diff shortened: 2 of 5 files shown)


cfallin:
Thanks for this; the semantics you've described make reasonable sense each in isolation, but I worry that this interface will be too confusing/complex for a user to invoke and have reasonable expectations for. Especially the transparent shift between one-flags-applies-to-all and flags-apply-pairwise is a mistake waiting to happen: forgetting some flag options could switch modes transparently and cause unexpected flags to apply to later engines.

I wonder if we could disambiguate intent with two different flags: `--engine-flags-common` and `--engine-flags`. Then there would be two canonical invocation idioms:

* `-e engine1.so --engine-flags flag1 -e engine2.so --engine-flags flag2 ...`

  (the arg-parser doesn't enforce the interleaving, but this would be a nice and clear way to write the command line); and

* `-e engine1.so -e engine2.so ... --engine-flags-common universal-flag`

What's more, they could *also* compose, which might be really useful: specify common flags once, but also add flags to each run individually.

This seems to me to be a minor change to your logic, and overall simplifies the semantics / makes them more understandable. What do you think?

fitzgen:
> This seems to me to be a minor change to your logic, and overall simplifies the semantics / makes them more understandable. What do you think?

I like this, will implement it shortly

fitzgen:
@cfallin ready for re-review!

cfallin (approved):
👍

PR #331, opened 2026-08-12
Add "Sum Total" rows to analyses of existing data, not just fresh benchmarks

fitzgen:
We were previously only reporting "Sum Total" rows for freshly recorded benchmarks, and not for `sightglass effect-size ...` or `sightglass summarize ...` analyses on already-recorded data. This commit makes it so that we always report the "Sum Total" row in both cases.

Also fixes a bug where we weren't totalling across processes, so each process got its own "Sum Total" row instead of accumulating them all together into one row.


fitzgen:
@posborne mind taking a look at this one?

posborne (approved):

posborne (approved):
I probably should have noticed that missing module in my review; looks good to me.

PR #332, opened 2026-08-12
Tune nearly all benchmarks to ~100M wasm instructions executed per iteration

fitzgen:
Note that this was not possible for a few of the benchmarks, so we did the best we could. See the changes to `README.md` files for details on each case.


fitzgen:
@cfallin gentle review ping here

benchmarks/libsodium/README.md
```diff
@@ -0,0 +1,115 @@
...
+| `pwhash_argon2id` | 5.92G | `scalarmult8` | 292.9M |
+| `pwhash_argon2i` | 5.20G | `sodium_utils` | 242.9M |
+| `ed25519_convert` | 4.17G | `auth7` | 228.2M |
+| `core3` | 1.64G | `aead_aegis256` | 151.1M |
+| `pwhash_scrypt_ll` | 1.40G | `secretbox7` | 140.3M |
+| `secretbox_easy2` | 1.18G | `aead_aegis128l` | 136.1M |
+| `stream` | 1.12G | `onetimeauth7` | 131.6M |
+| `stream2` | 1.03G |  |  |
+
+### No measurable work on Wasm (6)
```
cfallin, line 74: Should we just delete these benchmarks? "Does not do anything" (e.g. `return 0`) seems like an objective, easily defensible reason to do so...

cfallin (approved):
Sorry about the delay; this one fell off my queue somehow. Thanks!

benchmarks/libsodium/README.md
fitzgen: Will do in a follow up!

PR #333, opened 2026-08-24
Remove libsodium benchmarks that do no measurable work on Wasm

fitzgen:
@cfallin here is the promised follow up PR


cfallin (approved):
Thanks!

PR #334, opened 2026-09-23
Remove rust-precision dependency

theotherjimmy:
Since we only use the cpu counters from rust-precision, and it's stale I've elected to add support for s390x by removing the denpendency.


crates/recorder/src/measure/cycles.rs
```diff
@@ -1,23 +1,51 @@
...
+    time
+}
+
+#[inline]
+#[cfg(target_arch = "s390x")]
+fn cpucounter() -> u64 {
+    let time: u64;
+    unsafe { asm!("stck ({})", out(reg) &mut time); }
+    time
+}
```
fitzgen: Mind using https://doc.rust-lang.org/nightly/std/macro.cfg_select.html for these `cfg`s so that we get a decent compiler error message when none applies for new architectures? Thanks!

crates/recorder/src/measure/cycles.rs
```diff
@@ -33,17 +61,19 @@ impl CycleMeasure {
...
-
-        let end = precision.now();
-        let elapsed = end - self.0.take().expect("must call start before end");
+        let end = cpucounter();
+        let start = self.0.take().expect("must call start before end");
+        let elapsed = if end >= start {
+            end - start
+        } else {
+            end + start + 1
+        };
```
fitzgen: Do we want this arithmetic to using the wrapping methods?

fitzgen (commented):
Thanks! A suggestion and a question below

